PR #3, opened 2018-06-12
Add statusbar height to cover the notch on Xiaomi Mi 8

TadiT7:


phhusson:
you need to add treble-overlay-xiaomi-mi8 to overlay.mk
else, that's good for me

phhusson:
Thanks!
It would be nice to extract other useful values from stock framework-res like automatic backlight

phhusson:
Oh well, priority needed to be increased. I did in another commit, but that's a note for later overlays

PR #8, opened 2018-07-05
overlay: Apply padding for rounded corner

axm2:
https://github.com/LineageOS/android_device_oneplus_sdm845-common/commit/91ff5c04b44124fa90132ffe10ad004c6c7e87c8


phhusson:
When I tested it, it didn't work, nothing changed.
But I might have failed somewhere.
So do you confirm you tested this actual patch?

phhusson:
Well your patch is made on framewotk-res, not on systemui
So I'm pretty sure it doesn't work. 
But I'm putting that on my to-do list

PR #10, opened 2018-07-07
Add overlay for Sharp AQUOS S2

AndroPlus-org:


Sharp/S2/res/values/config.xml
```diff
@@ -0,0 +1,103 @@
...
+    <bool name="config_dozePulsePickup">true</bool>
+    <bool name="config_hotswapCapable">true</bool>
+    <bool name="config_intrusiveNotificationLed">true</bool>
+    <bool name="config_lidControlsSleep">true</bool>
+    <bool name="config_pinnerCameraApp">true</bool>
+    <bool name="config_powerDecoupleInteractiveModeFromDisplay">true</bool>
+    <bool name="config_setColorTransformAccelerated">true</bool>
+    <bool name="config_suspendWhenScreenOffDueToProximity">true</bool>
+    <bool name="config_sustainedPerformanceModeSupported">true</bool>
+    <bool name="config_useRoundIcon">false</bool>
```
phhusson: Nope, that's not device-specific but ROM-specific (also this parameter is obsolete)

Why did you add this? Does this come from stock ROM, or you actually want this?

Sharp/S2/res/values/config.xml
```diff
@@ -0,0 +1,103 @@
...
+    <bool name="config_wifi_background_scan_support">true</bool>
+    <bool name="config_wifi_batched_scan_supported">true</bool>
+    <bool name="config_wifi_dual_band_support">true</bool>
+    <bool name="skip_restoring_network_selection">true</bool>
+    <integer name="config_bluetooth_operating_voltage_mv">3300</integer>
+    <integer name="config_multiuserMaximumUsers">4</integer>
+    <integer name="config_wifi_wakeup_available">1</integer>
+    <integer name="config_backPanicBehavior">1</integer>
```
phhusson: nope, not device-specific either.
If you think this parameter is relevant feel free to open a PR on device/phh/treble's overlay

Sharp/S2/res/values/config.xml
```diff
@@ -0,0 +1,103 @@
...
+    </integer-array>
+
+    <bool name="config_bluetooth_hfp_inband_ringing_support">true</bool>
+    <bool name="config_bluetooth_le_peripheral_mode_supported">true</bool>
+    <bool name="config_dozeAfterScreenOff">true</bool>
+    <bool name="config_dozePulsePickup">true</bool>
+    <bool name="config_hotswapCapable">true</bool>
+    <bool name="config_intrusiveNotificationLed">true</bool>
+    <bool name="config_lidControlsSleep">true</bool>
+    <bool name="config_pinnerCameraApp">true</bool>
```
phhusson: Ok, so this one is technically device-specific, but only because it depends on the quantity of RAM of the device.

I plan on doing an overlay for >=2GB (or >2GB?) to include this, the other pin framework option, multiuser, and probably other stuff

Can you please move it (and multiuser) at the end of the XML, with a comment saying <-- big ram --> or something like that?

Sharp/S2/res/values/config.xml
```diff
@@ -0,0 +1,103 @@
...
+    <bool name="config_setColorTransformAccelerated">true</bool>
+    <bool name="config_suspendWhenScreenOffDueToProximity">true</bool>
+    <bool name="config_sustainedPerformanceModeSupported">true</bool>
+    <bool name="config_useRoundIcon">false</bool>
+    <bool name="config_wifi_background_scan_support">true</bool>
+    <bool name="config_wifi_batched_scan_supported">true</bool>
+    <bool name="config_wifi_dual_band_support">true</bool>
+    <bool name="skip_restoring_network_selection">true</bool>
+    <integer name="config_bluetooth_operating_voltage_mv">3300</integer>
+    <integer name="config_multiuserMaximumUsers">4</integer>
```
phhusson: cf config_pinnerCameraApp

AndroPlus-org:
Yes, they are copied from stock.
I cleaned up to only include pin and multiuser option.

phhusson:
Can you also include stock res/xml/power_profile.xml?
This should fix battery tracking in settings

AndroPlus-org:
OK, I added power profile.

phhusson:
Thanks!

PR #21, opened 2018-08-28
More Huawei overlays.

Dil3mm4:
Devices supported as of now:
-STF
-VTR
-ANE
-BND
-LLD
-PRA
-RNE
-WAS
-DUK


phhusson:
Priorities are broken, but I'll fix it in another commit

PR #23, opened 2018-09-01
Add svg paths for notch detection on P

TadiT7:


phhusson:
So that's dumb rectangles right?

TadiT7:
Yup xD I just wonder if OEMs would do any better. Anyway, it's invisible to the user without the config_fillMainBuiltInDisplayCutout flag and does what it is supposed to do. Some width values may just need polishing, because I had to base on renders in case of few devices.

Issue #25, opened 2018-09-15
ro.vendor.product.device is deprecated in API 28 vendors


argraur:
Starting with API 28 OEMs use ro.product.vendor.device for device name definition in /vendor/build.prop

Xiaomi Mi 8: 
https://github.com/TadiT7/xiaomi_dipper_dump/blob/8385f43b85cc3c4f055fff6289776374c4c5d6e9/vendor/build.prop#L18
Xiaomi Mi Mix 2S: https://github.com/TadiT7/xiaomi_polaris_dump/blob/d069dedee55a3fb62d003a3a30a7642cb1d2e295/vendor/build.prop#L18
Google Pixel 2: https://github.com/TadiT7/google_walleye_dump/blob/8033f01f57d9cf11b3b2686c51a4a3e9a4dffaef/vendor/build.prop#L18
OnePlus 6: https://github.com/TadiT7/oneplus_enchilada_dump/blob/d1462ce505d9e70d0288ed4359ef21b998e9edbe/vendor/build.prop#L10

argraur:
Sure sir

PR #26, opened 2018-09-16
Use build fingerprint check for Mi 8

argraur:
Old way doesn't work with Pie vendors.
Move to universal way of determining Mi 8
See this: https://github.com/phhusson/vendor_hardware_overlay/issues/25

Test: make -j64 treble-overlay-xiaomi-mi8; use 9.0 vendor; copy overlay to /system/overlay/; ensure overlay works

Signed-off-by: argraur <argraur@gmail.com>


linuxct:
Pretty cool verification, will add it to cheryl too, thanks

PR #28, opened 2018-09-26
Add an overlay to enable always on display

NeonDragon1909:


phhusson:
Please have android:priority value be unique in vendor/hardware_overlay.
Except for that, good for me.

How do you plan to set the property?
I've been thinking of adding a toggle in Treble app
Or perhaps just have it always enabled, but always on disabled by default?

NeonDragon1909:
I'll create a new pr with a few more improvements. I wasn't thinking really well when i made the changes directly to the master branch and its making things a bit difficult

PR #29, opened 2018-09-26
aod overlay and misc. improvements

NeonDragon1909:


NeonDragon1909:
I opted for having the overlay always enabled. I'm not sure of how to make always on display disabled by default. The closest i could find is config_dozeAfterScreenOffByDefault but I'm not sure that's what I'm looking for.

NeonDragon1909:
I opted for having the overlay always enabled. I'm not sure of how to make always on display disabled by default. The closest i could find is config_dozeAfterScreenOffByDefault but I'm not sure that's what I'm looking for.

PR #31, opened 2018-10-07
Add resources for Xiaomi Mi 8 SE

Asureus:
fix phhusson/treble_experimentations#148

Additionally, for AOD we need to set "doze.display.supported=true"

Shall I add it to rw-system.sh?
(stock MIUI did it in /system/build.prop)


AlonsoK28:
**This fix auto-brightness option?**

phhusson:
SystemUI's code is:
mDisplayStateSupported = SystemProperties.getBoolean("doze.display.supported",
                res.getBoolean(R.bool.doze_display_state_supported))

So instead of setting the property, I'd rather do an overlay for it

PR #32, opened 2018-10-11
Add resources for some xiaomi (longcheer's odm) devices

suwakowww:
I think it can fix 
phhusson/treble_experimentations#226 and phhusson/treble_experimentations#229
for:
- wayne (tested and working)
- jasmine (blind-patched)
- sakura (blind-patched)
- daisy (blind-patched)
- nitrogen (blind-patched)
- whyred (blind-patched)

PR #33, opened 2018-10-17
added resource files for Moto G6

Wabo0:
I mostly copied the Moto G6 Plus files and changed the values which differed from the ones in my stock image `framework-res.apk`. 
Automatic brightness and battery usage stats (phhusson/treble_experimentations#249) now work for me.

### DIFF
--- a/Moto/G6/AndroidManifest.xml
+++ b/Moto/G6/AndroidManifest.xml
@@ -5,6 +5,6 @@
         <overlay android:targetPackage="android"
                 android:requiredSystemPropertyName="ro.vendor.product.name"
                 android:requiredSystemPropertyValue="ali_retail"
-		android:priority="21"
+		android:priority="43"
 		android:isStatic="true" />
 </manifest>
--- a/Moto/G6/AndroidManifest.xml
+++ b/Moto/G6/AndroidManifest.xml
@@ -1,0 +1,10 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+        package="me.phh.treble.overlay.moto.g6"
+        android:versionCode="1"
+        android:versionName="1.0">
+        <overlay android:targetPackage="android"
+                android:requiredSystemPropertyName="ro.vendor.product.name"
+                android:requiredSystemPropertyValue="ali_retail"
+		android:priority="21"
+		android:isStatic="true" />
+</manifest>
--- a/Moto/G6/AndroidManifest.xml
+++ b/Moto/G6/AndroidManifest.xml
@@ -3,8 +3,8 @@
         android:versionCode="1"
         android:versionName="1.0">
         <overlay android:targetPackage="android"
-                android:requiredSystemPropertyName="ro.vendor.product.name"
-                android:requiredSystemPropertyValue="ali_retail"
+                android:requiredSystemPropertyName="ro.vendor.build.fingerprint"
+                android:requiredSystemPropertyValue="+motorola/ali_*"
 		android:priority="43"
 		android:isStatic="true" />
 </manifest>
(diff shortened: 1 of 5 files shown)


Moto/G6/res/values/config.xml
```diff
@@ -10,7 +10,7 @@
 	<integer name="config_brightness_ramp_rate_fast">200</integer>
 	<integer name="config_brightness_ramp_rate_slow">40</integer>
 	<bool name="config_setColorTransformAccelerated">true</bool>
-	<bool name="config_nightDisplayAvailable">false</bool>
```
phhusson: By default, config_nightDisplayAvailable is set to config_setColorTransformAccelerated
So simply removing this line will do

Moto/G6/AndroidManifest.xml
```diff
@@ -0,0 +1,10 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+        package="me.phh.treble.overlay.moto.g6"
+        android:versionCode="1"
+        android:versionName="1.0">
+        <overlay android:targetPackage="android"
+                android:requiredSystemPropertyName="ro.vendor.product.name"
+                android:requiredSystemPropertyValue="ali_retail"
```
phhusson: All Moto G6 are called ali_retail, there won't be any variants, like ali_verizon, or ali_india or such stuff?

Moto/G6/AndroidManifest.xml
```diff
@@ -0,0 +1,10 @@
...
+		android:priority="21"
```
phhusson: Priorities need to be unique. Please use 43.

Moto/G6/AndroidManifest.xml
Wabo0: I don't know if there are other version. Would something like `"ali_*"` work or do I have to use some entirely different SystemProperty?

Wabo0: I don't know if there are other version. Would something like `"ali_*"` work or do I have to use some entirely different SystemProperty?

Moto/G6/res/values/config.xml
Wabo0: Done. (dc8bf86)

Moto/G6/AndroidManifest.xml
Wabo0: Done. (5ac58de)

Moto/G6/AndroidManifest.xml
phhusson: Also please use fingerprint instead of ro.vendor.product to be pie-vendor proof.
And yes use ali_*
See https://github.com/phhusson/vendor_hardware_overlay/pull/32/commits/97c46901fbd28002d3d028f9575e285bfb1a190c as an example

Wabo0: Done. (5ec4807)
(Though I haven't tested it yet. I'll note it here once I get around to that.)

phhusson:
Could you confirm the fingerprint check works?

I've noticed that on evert it's not the case, so I'm expecting the same for ali ( https://github.com/phhusson/vendor_hardware_overlay/commit/53b36d20d6d670ec7cda36ec601db39fb4f73389#diff-fc50d313235230bcf8e47d962a526249R5 )

Wabo0:
> Could you confirm the fingerprint check works?
> 
> I've noticed that on evert it's not the case, so I'm expecting the same for ali ( [53b36d2#diff-fc50d313235230bcf8e47d962a526249R5](https://github.com/phhusson/vendor_hardware_overlay/commit/53b36d20d6d670ec7cda36ec601db39fb4f73389#diff-fc50d313235230bcf8e47d962a526249R5) )

`$ adb shell getprop ro.vendor.build.fingerprint`
returns 
`motorola/ali_n/ali_n:8.0.0/OPS27.82-41/50:user/release-keys`

So I think it should work. Still downloading the android sources again for a real build/test on device, though that might take a few more days.

phhusson:
You don't need a full build to tests overlays
You can just run build/build.sh
But anyway that's enough for me
Thanks

PR #37, opened 2018-11-08
Support rounded corners, change some values

Dekamir:
Disable all rotations because we have a notch.
Add support for rounded corners.

### DIFF
--- a/Huawei/hi6250/ANE/res/values/config.xml
+++ b/Huawei/hi6250/ANE/res/values/config.xml
@@ -26,7 +26,7 @@
 <resources xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
     <!-- If true, the screen can be rotated via the accelerometer in all 4
          rotations as the default behavior. -->
-    <bool name="config_allowAllRotations">true</bool>  
+    <bool name="config_allowAllRotations">false</bool>  
     
 	<!-- Array of light sensor LUX values to define our levels for auto backlight brightness support.
          The N entries of this array define N + 1 control points as follows:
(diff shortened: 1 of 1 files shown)


Huawei/hi6250/ANE/res/values/config.xml
```diff
@@ -96,7 +96,7 @@
             1 - Toggle theater mode setting
             2 - Brightness boost
     -->
-    <integer name="config_doublePressOnPowerBehavior">2</integer>
+    <integer name="config_doublePressOnPowerBehavior">0</integer>
```
phhusson: Uhhhhh yeah, wtf was that. @Dil3mm4 ?

Huawei/hi6250/ANE/res/values/config.xml
```diff
@@ -227,6 +227,10 @@
 
     <!-- When true use the linux /dev/input/event subsystem to detect the switch changes
          on the headphone/microphone jack. When false use the older uevent framework. -->
+	
+    <!-- When set, it narrows the system user interface to fix problems with aspect ratio and
+         rounded corners. -->
+    <dimen name="rounded_corner_content_padding">58px</dimen>
```
phhusson: This is a SystemUI value, not a framework-res value, so it will not work.
But since Pie (Pie DR1?), you can set rounded_corner_radius in framework-res instead
So change to that. (If you tell me it really works this way, please tell, but I would have hard time believing you)

Dekamir: _Tadi told me to copy paste this._
I mean, you can just make Samsung's SystemUI work on P20 lite.
Or we can create another System UI overlay for it.
Enabling overlay manually works fine.
My git and Android information ends here, sorry. :(

Huawei/hi6250/ANE/res/values/config.xml
Dil3mm4: Dumbness that will be rebased along brightness values for Huawei devices.
My bad.

PR #40, opened 2018-11-12
Improve brightness levels for Huawei and remove some messy value.

Dil3mm4:
Values are now more granular and they look better on mid-brightness settings.
Double tap behaviour bool it's now set to 0.


TadiT7:
https://android.googlesource.com/platform/frameworks/base/+/pie-qpr1-release/core/res/res/values/config.xml#1047
remove this too

PR #41, opened 2018-11-19
Adjust Mi 8SE statusbar and rounded corners

TadiT7:


TadiT7:
Yeah, CI needs to include that, as we already build with Pie values normally, like status_bar_height_portrait

PR #48, opened 2018-12-10
Update notch support

TadiT7:


TadiT7:
uh xD

TadiT7:
Ping ping @phhusson

phhusson:
So config_mainBuiltInDisplayCutout is the real hardware thing, and was correct in your initial commit
but to "expand touchable area", you decided to make it bigger, and now you decide to set it to hardware size?
Or initial values of status_bar_height and config_mainBuiltInDisplayCutout were simply cherry-picked from OEM?

TadiT7:
Tested on actual tulip. Looks well, but the huge notch lets us to fit only one notification icon. Or rather AOSP won't let us keep them denser. Treat as finished.

PR #50, opened 2018-12-23
Update overlay for Mi MIX 3

AndroPlus-org:
* Imported brightness/doze config from 8.12.13
* config_displayBlanksAfterDoze is true in stock, but sometimes status bar is shown in AOD, so disable it for now.
* Enable rounded corner padding

Without enabling SystemUI overlay, AOD will disappear sometimes.
https://github.com/phhusson/treble_app/pull/8


phhusson:
Uh autobacklight works?
Sounds like config_autoBrightnessLcdBacklightValues values goes up to 2047, which doesn't sound right on AOSP

TadiT7:
> Uh autobacklight works?
> Sounds like config_autoBrightnessLcdBacklightValues values goes up to 2047, which doesn't sound right on AOSP
> 
> CAF has support for > 8 bits brightness, but AOSP doesn't

I'll just ping @argraur to be sure

phhusson:
Merged, but I removed brightness part for the moment:
https://github.com/phhusson/vendor_hardware_overlay/commit/4ea7508cbabc61c6487d51f09d687e69af4aca72

Feel free to keep updating this PR, I'll take care of the merge.
But please keep my updated about this auto backlight issue.

AndroPlus-org:
@phhusson
I updated the brightness config and tested on my mix 3.
I also added a fix for AOD disappearing / not entering into deep sleep issue.

phhusson:
```
     <bool name="doze_suspend_display_state_supported">false</bool> 
```
This sounds sad. I merged your PR in without this setting for the moment I'll test it a bit.

Thanks for the whole PR though

PR #53, opened 2018-12-28
Add overlay for oneplus 6t

bengris32:


phhusson:
Sorry for the noise, but you can reenable config_lightSensorWarmupTime if you want, I'll ignore that error, which isn't a real one.

bengris32:
> Sorry for the noise, but you can reenable config_lightSensorWarmupTime if you want, I'll ignore that error, which isn't a real one.

ok no problem

PR #58, opened 2019-01-21
Add initial notch overlay for Xiaomi Mi Play, Redmi Note 7, Mi 9 and Mi 9 Lite(?)

TadiT7:
Data of the two latter devices was extracted from a system app. Their existence and codenames are certain, so eventually they'll need a polishment one day, but we'll be prepared.


phhusson:
You know, there is more to a device than just notches ;-)

TadiT7:
> You know, there is more to a device than just notches ;-)

My thoughts exactly, to be honest. It's due to laziness and a motivation to play safe, as OEMs are oftentimes less trustable than AOSP defaults (like that 0 minimal value of brightness on some meme phone). Either way, it all needs to be verified by a real, sane device user, but that's rare.

PR #60, opened 2019-01-22
Razer Cheryl Overlay (Clean)

linuxct:


TadiT7:
> Already pushed commits can be squashed????? @TadiT7

Yes, you can just force push after that, since it's just a pull request branch.

linuxct:
@TadiT7 done, thanks

PR #67, opened 2019-02-04
 Dehardcode status_bar_height for Pie

TadiT7:
It caused issues with other overlays, like NoCutout. Still needed for Oreo.


phhusson:
github seems totally lost with this pr, I am too. It's all good now?

PR #68, opened 2019-02-05
beryllium: Update notch cutout configs from P beta

amog787:
* Extracted from POCOF1Global_8.11.2, framework-res.apk
* Xiaomi's notch cutout config was configured oddly, redo the pathmap so that the pathmap actually starts from origin (0,0)
* Round-off the values to 1dp

### DIFF
--- a/Xiaomi/PocoF1/res/values/notch.xml
+++ b/Xiaomi/PocoF1/res/values/notch.xml
@@ -18,10 +18,19 @@
 -->
 <resources>
     <!-- Height of the status bar in portrait -->
-    <dimen name="status_bar_height_portrait">90px</dimen>
+    <dimen name="status_bar_height_portrait">87px</dimen>
     <!-- Height of the status bar -->
     <dimen name="status_bar_height">90px</dimen>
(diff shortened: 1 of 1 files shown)


amog787:
> The thing is, have you tested your change? Negative starting values were officially shown as an example in the AOSP repo, so they are used in a right way.
> Also keep in mind the amount of visible statusbar elements, because better curves eat more space and are completely useless if we aren't rendering them :/

Yep tested.
The problem with the previous values is that there is a rectangular box(visible) instead of the curved notch. This value fixes it. :3

TadiT7:
> > The thing is, have you tested your change? Negative starting values were officially shown as an example in the AOSP repo, so they are used in a right way.
> > Also keep in mind the amount of visible statusbar elements, because better curves eat more space and are completely useless if we aren't rendering them :/
> 
> Yep tested.
> The problem with the previous values is that there is a rectangular box(visible) instead of the curved notch. This value fixes it. :3

Sorry, but you aren't using completely stock vendor or your system is modified. `config_fillMainBuiltInDisplayCutout` isn't set to true by us.

PR #71, opened 2019-02-09
Update notch.xml

SmartRu:
Chenge the status_bar_height


SmartRu:
I got this height from stock room and have tested it... Also I think the other files also have sth. wrong

phhusson:
Did you check who authored previous value?
Ping them here and discuss it

TadiT7:
It makes no sense to alter landscape statusbar height from AOSP defaults

PR #72, opened 2019-02-09
patch the power_profile.xml

SmartRu:
Get this files from stock framework-res. apk


sellerie98:
This did not fix the issue as battery stats are still not appearing.

PR #73, opened 2019-02-11
Added overlay for Zenfone Live L1 (X00RD)

v555574:


v555574:
> I doubt this works properly, values in also can't go higher than 255

All values where copied from OEM Rom.

TadiT7:
"Oreo" part in the name is confusing. If there's an important difference between ZF Live L1 variants, then I recommend to add "_X00RD" instead

v555574:
> "Oreo" part in the name is confusing. If there's an important difference between ZF Live L1 variants, then I recommend to add "_X00RD" instead

There is 1/8 version with Android Go, and they share codename (model no.) for some reason. Changed to model name, it seems to be unique.

TadiT7:
But vendor fingerprint remains the same between both variants? If we differentiate these devices, we need to make sure that we actually do it in the overlay's code.

v555574:
> But vendor fingerprint remains the same between both variants? If we differentiate these devices, we need to make sure that we actually do it in the overlay's code.

I don't have access to Android Go version, but i think it's '2' in 'ASUS_X00R_2' that is differentiates Oreo version. I have that as my property value.

TadiT7:
Download the firmware on the Asus website then.

phhusson:
Did you test this version?
You did a rule of three?

v555574:
Tested, working just fine.
Putting values calculated by rule of three resulted in very dim screen. These are evenly distributed values, looks closer to stock ROM.

phhusson:
Ok, thanks!

PR #77, opened 2019-02-16
Actually fix power profile on LG G7 ThinQ

sellerie98:
As the PR #72 didn't fix the issue for some reason I investigated further and realized that the file was in the wrong folder. I tested it and it is working on my device now.


phhusson:
Arf...
Please remove the other file in that PR

sellerie98:
I don't really understand. I mean - I moved it, not copied it. The other file (I assume that you mean the file in the old, wrong place, doesnt exist anymore.

phhusson:
ah yes sorry

PR #78, opened 2019-02-17
Enable the navigation bar for Mi A2

jackwakefield:
This is currently missing for Pie.

### DIFF
--- a/Xiaomi/MiA2/res/values/config.xml
+++ b/Xiaomi/MiA2/res/values/config.xml
@@ -46,6 +46,8 @@
     <bool name="config_dozeAlwaysOnDisplayAvailable">false</bool>
     <bool name="config_displayBlanksAfterDoze">false</bool>
     <bool name="config_powerDecoupleAutoSuspendModeFromDisplay">false</bool>
+    <bool name="config_showNavigationBar">true</bool>
+    <bool name="config_supportSystemNavigationKeys">true</bool>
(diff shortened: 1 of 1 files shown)


TadiT7:
What https://github.com/phhusson/vendor_hardware_overlay/blob/08c4ef3d9c851bc27e625be3a685482c434d975b/Xiaomi/MiA2/res/values/config.xml#L32 has to do with navbar?

jackwakefield:
Whoops, that was meant to be removed (and is now).

phhusson:
config_supportSystemNavigationKeys has nothing to do with navbar (yeah thanks Google for that helpful naming.)
This is fingerprint sensor gestures (swipe up/down on fingerprint sensor to show/hide notifications)

Is this intended?

jackwakefield:
Ah, I had included config_supportSystemNavigationKeys because I thought it was to do with the navbar (thanks for the heads up), so it could be removed at least from this PR, but it is in the Pie ROM's config.

On the other hand, I haven't yet had any luck getting the fingerprint reader to work on the Mi A2, but I haven't investigated that yet and it could just be an issue on my end.

I'm happy to take your advice on whether to remove it or separate it out into a new PR.

phhusson:
merged in https://github.com/phhusson/vendor_hardware_overlay/commit/1643f1d9fd13b1e37feeadca558f08dbc11b19c5

PR #79, opened 2019-02-18
Overlay: OnePlus 6/6T: properly check for device.

Herna1994:
With all my respects, the current way of detecting if it´s
a OnePlus 6 or 6T is bit bad.

Switch to ro.boot.project_name, as we have done in Paranoid Android
for the last years, easiest way.

Signed-off-by: Hernán Castañón <herna@paranoidandroid.co>

### DIFF
--- a/OnePlus/OP6/AndroidManifest.xml
+++ b/OnePlus/OP6/AndroidManifest.xml
@@ -3,8 +3,8 @@
         android:versionCode="1"
         android:versionName="1.0">
         <overlay android:targetPackage="android"
-                android:requiredSystemPropertyName="ro.vendor.build.fingerprint"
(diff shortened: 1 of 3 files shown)


TadiT7:
ro.boot.project_name is fully device specific?

Herna1994:
Yes, it´s OnePlus specific. Has been there since their first smartphone.

TadiT7:
Pros:
requires no matching patch
can be used with altered fingerprint
Cons:
is more likely to change or disappear, compared to the fingerprint string

rest is up to @phhusson

Herna1994:
It won't disappear. It's been 8 devices, they didn't get rid of it.

PR #82, opened 2019-02-22
Add brightness curve for AlwaysOnDisplay

sellerie98:
This commit add a seperate brightness curve for the AlwaysOnDisplay for the LG G7 ThinQ. This is necessary for the feature to make sense on the device because it has an IPS screen. and the backlight needs to be turned on for it, but not as bright as it needs to be for usual operation because the screen has white sub-pixels.

### DIFF
--- a/LG/G7/res/values/config.xml
+++ b/LG/G7/res/values/config.xml
@@ -32,6 +32,16 @@
         <item>1600</item>
         <item>3100</item>
     </integer-array>
+    <array name="config_autoBrightnessDisplayValuesNits">
+        <item>8.2</item>
+        <item>23.3</item>
+        <item>122.28</item>
+        <item>144.4</item>
+        <item>189.4</item>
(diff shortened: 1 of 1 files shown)


phhusson:
Your description is wrong, this isn't separate for Always On Display, it gets used by standard auto brightness as well.
Though with proper nits value, it makes sense that AOD backlight has a better behaviour

Merged in https://github.com/phhusson/vendor_hardware_overlay/commit/f1492d1c2a8a0dc858cad69aa82cae8612bdbc76

sellerie98:
Can I find docu about this somewhere? I'd love to know what the difference in these settings is

phhusson:
It's documented along with default values:
https://android.googlesource.com/platform/frameworks/base/+/refs/tags/android-9.0.0_r34/core/res/res/values/config.xml#1407

PR #84, opened 2019-02-25
Final fix for ZA550KL

v555574:
Notification LED working, everything else working as on stock, fully tested.


TadiT7:
You can drop first commit since we do this here
https://github.com/phhusson/device_phh_treble/blob/android-9.0/overlay/frameworks/base/core/res/res/values/config.xml#L21

TadiT7:
So the whole commit seems to do nothing

TadiT7:
Aaah, sorry, just delete the line then :)

TadiT7:
Or better, you can uncomment it, adding the info about the user level fix, so it'll stand out

PR #85, opened 2019-03-02
Add Motorola One (deen) notch overlay

milankragujevic:


TadiT7:
One thing to remember: in the pie branch, we need to remove the plain `status_bar_height` item, because it's only needed for Oreo and causes glitches on Pie. Like this: https://github.com/phhusson/vendor_hardware_overlay/commit/e7b2645ffb6a0daf70ed16fb97b607f9ffb82b16

phhusson:
Merged in https://github.com/phhusson/vendor_hardware_overlay/commit/e426db35989b7ca86d7318014450e08c45ee80fe

PR #87, opened 2019-03-05
Fix led notifications lenovo s5

Mistrick:
Seems i wrongly understood this param and broke led


phhusson:
What if you set this value to true? Is led still fixed in that case?

Mistrick:
If it false we lose any LED settings in app notification settings.
I test LED with some apps and saw only blue color(if it true).

TadiT7:
> What if you set this value to true? Is led still fixed in that case?

We set it to true by default, so deleting is the right action in this pr, by the way.

Mistrick:
merge?

PR #92, opened 2019-03-19
Introduce XZ3 overlays for Pie

Paulbouchara:
Those are the needed overlays for XZ3 curved display and cutout configuration (round corners)
Tested and working on H9436 ( XZ3 Dual SIM)
The actual tests errors are because of Pie difference with Oreo, hence XZ3 got out with Pie and does need some of those overlays, so we don't care about those errors


phhusson:
Merged in https://github.com/phhusson/vendor_hardware_overlay/commit/bd64c0a20fd04e929a6f0271481800d357befd71

PR #110, opened 2019-06-03
Add resources for Xiaomi Redmi K20 Pro (raphael)

AndyCGYan:
Numerical values taken directly from device; bool values set based on other devices (Mi 8, Mi Mix 3).
I can confirm that this at least fixes auto brightness, not sure what else.
Since this is my first PR ever on GitHub, if I made any mistakes, or if there's something wrong within the changes themselves, please let me know.

### DIFF
--- a/Xiaomi/RedmiK20Pro/Android.mk
+++ b/Xiaomi/RedmiK20Pro/Android.mk
@@ -1,0 +1,8 @@
+LOCAL_PATH := $(call my-dir)
+include $(CLEAR_VARS)
+LOCAL_MODULE_TAGS := optional
+LOCAL_PACKAGE_NAME := treble-overlay-xiaomi-redmik20pro
+LOCAL_MODULE_PATH := $(TARGET_OUT)/overlay
+LOCAL_IS_RUNTIME_RESOURCE_OVERLAY := true
+LOCAL_PRIVATE_PLATFORM_APIS := true
+include $(BUILD_PACKAGE)
--- a/Xiaomi/RedmiK20Pro/AndroidManifest.xml
+++ b/Xiaomi/RedmiK20Pro/AndroidManifest.xml
@@ -1,0 +1,10 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+        package="me.phh.treble.overlay.xiaomi.redmik20pro"
+        android:versionCode="1"
+        android:versionName="1.0">
+        <overlay android:targetPackage="android"
+                android:requiredSystemPropertyName="ro.vendor.build.fingerprint"
+                android:requiredSystemPropertyValue="+*iaomi/raphael*"
+		android:priority="111"
+		android:isStatic="true" />
+</manifest>
--- a/Xiaomi/RedmiK20Pro/res/values/config.xml
+++ b/Xiaomi/RedmiK20Pro/res/values/config.xml
@@ -195,7 +195,7 @@
 	<bool name="config_device_vt_available">true</bool>
 	<bool name="config_device_wfc_ims_available">true</bool>
 	<bool name="config_hotswapCapable">true</bool>
-	<bool name="config_intrusiveNotificationLed">false</bool>
+	<bool name="config_intrusiveNotificationLed">true</bool>
 	<bool name="config_lidControlsSleep">true</bool>
 	<bool name="config_wifiDisplaySupportsProtectedBuffers">true</bool>
 	<bool name="config_wifi_background_scan_support">true</bool>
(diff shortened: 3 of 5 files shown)


phhusson:
CI not happy
```
2019-06-03T12:51:20.8472323Z F: ./Xiaomi/RedmiK20Pro/res/values/config.xml: defines a non-existing attribute config_dozeAlwaysOnEnabled
```

AndyCGYan:
> CI not happy
> 
> ```
> 2019-06-03T12:51:20.8472323Z F: ./Xiaomi/RedmiK20Pro/res/values/config.xml: defines a non-existing attribute config_dozeAlwaysOnEnabled
> ```

Jeez, build.sh didn't complain about this.
But why though? Isn't this an existing prop in AOSP?

phhusson:
That's tests/tests.sh

This values does exist in AOSP, but there is no point in overlay-ing it,
it's true by default.
The use of this value is for ROMs that consider than AOD should be disabled
by default, it doesn't change the capability of the firmware to optionally
do AOD

phhusson:
It's not build that was broken, but tests :p

AndyCGYan:
> It's not build that was broken, but tests :p

I hate GitHub Desktop for not letting me amend commits >:(

AndyCGYan:
Er. what? Is it a requirement to use continuous priority numbers? But I sorted the existing list and I swear I saw not-continuous ones...

phhusson:
Nope
Issue is ```
F: ./Moto/G7Power/AndroidManifest.xml: ro.vendor.product.* is deprecated. Please use ro.vendor.build.fingerprint
```
so that's not you (I need to fix it though. I thought i already did).

Let's say CI is green.

Now we're waiting for @TadiT7 approval who has much better eyes than me for overlays.

AndyCGYan:
> Nope
> ...
> so that's not you (I need to fix it though. I thought i already did).
> 
> Let's say CI is green.
> 
> Now we're waiting for @TadiT7 approval who has much better eyes than me for overlays.

Thanks, waiting.
P.S. Fix the comment lol

TadiT7:
`<dimen name="status_bar_height_portrait">74px</dimen>` will do.

AndyCGYan:
> Seems clean, but I recommend to use a bigger statusbar, since it's quite a long device and OEMs now tend to make them bigger on such, even without any kind of a notch.
> Let's say 32dp would be nice, or at least the Google's 28dp.

Gotcha. Seems I need to create a separate SystemUI overlay for it.
28/32dp doesn't sound too drastic compared to the default 24dp, though I'll have to see it for myself, not a fan of too-big statusbar after using some notched devices.
Also, mind if I add `rounded_corner_content_padding` as well?

TadiT7:
> > Seems clean, but I recommend to use a bigger statusbar, since it's quite a long device and OEMs now tend to make them bigger on such, even without any kind of a notch.
> > Let's say 32dp would be nice, or at least the Google's 28dp.
> 
> Gotcha. Seems I need to create a separate SystemUI overlay for it.
> 28/32dp doesn't sound too drastic compared to the default 24dp, though I'll have to see it for myself, not a fan of too-big statusbar after using some notched devices.
> Also, mind if I add `rounded_corner_content_padding` as well?

Actually, statusbar size is the fwb part, while rounded corner content padding is only available through SystemUI. But 28dp will be fine, it doesn't look "notched" :)

AndyCGYan:
@TadiT7 Is it okay if I fix up (amend) the commit history and force push? Will that affect how the PR works?

AndyCGYan:
@TadiT7 Can you inspect the newly added SystemUI overlay please? It doesn't seem to get applied, at least on LOS.
It's possible though that it's an LOS-specific bug, in which case I'll just leave the overlay here and manually apply the padding on my device via ADB command.

phhusson:
SystemUI overlays need to be explicitly enabled, usually from treble app
This is annoying, so I simply added a parameter in treble settings so the
user can set its own rounded corners value.
Though if you do the relevant MR to treble app I can still merge it

AndyCGYan:
> SystemUI overlays need to be explicitly enabled, usually from treble app This is annoying, so I simply added a parameter in treble settings so the user can set its own rounded corners value. Though if you do the relevant MR to treble app I can still merge it

That sucks. Guess I'll just undo the SystemUI overlay.
Also, as I asked Tadi as well, can I do force push to correct commit history and such? Will it mess up this PR?

phhusson:
Yeah if you want

AndyCGYan:
Ugh, I just realized that I made changes on `master` instead of `pie`...

phhusson:
Yeah that works don't worry, I rebase pie on top of master for every
release.
(Except this month because I failed)

AndyCGYan:
> Yeah that works don't worry, I rebase pie on top of master for every release.

So I should actually switch the base back to `master`?

phhusson:
Both work

AndyCGYan:
OK, waiting on merge/rebase.

TadiT7:
px is the right way because it stays the same.

AndyCGYan:
> px is the right way because it stays the same.

Per PHH's reply I just ditched the SystemUI overlay altogether. Thanks for letting know though.

PR #111, opened 2019-06-05
Changed Prop for Redmi Note 7 Overlay

darkestDark:
Xiaomi uses the prop ro.vendor.build.fingerprint_real for vendor fingerprint.

Redmi Note 7 uses the MiMix2 overlay, because on the prop ro.vendor.build.fingerprint a polaris device is shown in /vendor/build.prop.

### DIFF
--- a/Xiaomi/MiMix2S/AndroidManifest.xml
+++ b/Xiaomi/MiMix2S/AndroidManifest.xml
@@ -3,7 +3,7 @@
         android:versionCode="1"
         android:versionName="1.0">
         <overlay android:targetPackage="android"
-                android:requiredSystemPropertyName="ro.vendor.build.fingerprint"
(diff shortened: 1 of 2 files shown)


phhusson:
To the best of my knowledge, this is a xiaomi.eu thingy which makes "Play
Store"-enabled china firmwares
Those are NOT firmware from Xiaomi

Le mar. 4 juin 2019 à 22:06, darkestDark <notifications@github.com> a
écrit :

darkestDark:
i see and read the topic... i think its okay that only play certificated devices ootb are supported. so i will went back to miui global stable. but i think nearly all users, which tested are dssi are on xiaomi.eu image. so they will affected by the bugs which are solved here by vendor fingerprint reading. 

- closed -

phhusson:
Fair enough. I'll try to add a warning in treble app
Issue opened at https://github.com/phhusson/treble_app/issues/16

PR #114, opened 2019-06-12
Fixed adaptive Brightness / Overlay applying on Redmi Note 7

darkestDark:


Xiaomi/RedmiNote7/res/values/notch.xml
```diff
@@ -18,8 +18,8 @@
 -->
 <resources>
     <!-- Height of the status bar in portrait -->
-    <dimen name="status_bar_height_portrait">89px</dimen>
```
phhusson: Uh nope, px are better than do (so that it works when user changes dpi)

darkestDark: no concerns... not looked so far.

darkestDark:
finally added the sunset and sunrise time... they werent include.
with vendor fingerprint the overlay wont be applied, idk why. the fingerprint is xiaomi stock vendor. have checked many different ones... no way. the vendor device works ootb.

phhusson:
I don't understand your last commit.
Those values are original AOSP values, there is no point in changing them...?

Also, this isn't hardware specific, this doesn't belong to device overlay.
You can open a MR in https://github.com/phhusson/device_phh_treble/tree/android-9.0/overlay/frameworks/base/core/res/res if you believe this should be changed in this AOSP ROM.

darkestDark:
okay...

PR #117, opened 2019-06-13
overlay: G7Power: correct priority

JarlPenguin:
* to pass tests


phhusson:
Thanks

phhusson:
But that's not the issue...

JarlPenguin:
@phhusson hm, it seemed like it to me, but I guess I'll just close this then?

PR #119, opened 2019-06-23
Add Addison (Moto Z Play) Overlay

Klozz:
Signed-off-by: klozz <klozz@vivaldi.net>


TadiT7:
It's not a device with official Treble implementation, everything can and should be solved in vendor. Speaking clearly, the purpose of this repo is to fix after OEMs who didn't include a proper overlay.
Thanks for the contribution tho, I'll add this info in the overlay guide.

PR #122, opened 2019-07-16
Enable config_intrusiveNotificationLed for Mi8 family

AndyCGYan:
...why the heck were these disabled in the first place?
Tested working on a Mi 8 SE.


phhusson:
Sorry about that

suwakowww:
These values pick from stock rom (based on 8.1) dumps, if it just a function, feel free to enable it.

phhusson:
Pushed in https://github.com/phhusson/vendor_hardware_overlay/commit/f64f2e164de749527cc329e66c6a71f817a1c284

PR #125, opened 2019-07-22
Overlay for Xiaomi Redmi 7 (Codename "onc") 

SebastianApel:
Dear Pierre-Hugues,
please find attached a pull request for the two commits: 
1) Add new overlay for Redmi 7 (Codename "onc") 
2) Add key "quick_qs_offset_height" to tests/knownKeys  (used in the Redmi 7 overlay)
Please let me know if you have any comments / change requests...
Kind regards,
Sebastian


TadiT7:
https://source.android.com/devices/tech/display/display-cutouts#implementation
Are you sure that 48dp is smaller than 80px on the onc?

SebastianApel:
Hi @TadiT7, I'm not sure if I understand your question correctly. To clarify:
- on the page you referenced, it says for "status_bar_height_portrait": "In most devices, this defaults to 24dp.". This is in line with the source code (https://android.googlesource.com/platform/frameworks/base/+/master/core/res/res/values/dimens.xml) and also with the behaviour I observed on my device  (statusbar height smaller than cutout, using Resurrection Remix GSI). 

- Where do the "48dp" you are talking about come from? (could it be that you are talking about the example on the page you referenced)?

- Regarding your question: on the device, "80px" is roughly "48dp". In my tests that resulted in a status bar higher than the cutout of 50px. I purposefully chose the higher value, because that moves the text in the status bar lower (so it is not hidden by the rounded corners)

Let me know if that resolves your question (or if you have other questions)...

TadiT7:
> * Where do the "48dp" you are talking about come from? (could it be that you are talking about the _example_ on the page you referenced)?

https://source.android.com/devices/tech/display/display-cutouts#implementation
"In values-land, set to status_bar_height_landscape, and in portrait set to either the default of 48dp, or the height of the cutout, whichever is larger. Can optionally be taller than the cutout if desired."
> * Regarding your question: on the device, "80px" is roughly "48dp". In my tests that resulted in a status bar higher than the cutout of 50px. I purposefully chose the higher value, because that moves the text in the status bar lower (so it is not hidden by the rounded corners)

What Google says is that you should have the `quick_qs_offset_height` value set at minimum 48dp. If it's roughly the same as 80px on onc and especially considering that you already chose a statusbar height higher than actual notch, this value should be left alone (deleted from the overlay), otherwise it'll look weird with any different dpi.

SebastianApel:
@TadiT7: Thanks for your reply! Will you or @phhusson merge the pull request?

Xiaomi/Redmi7onc/AndroidManifest.xml
```diff
@@ -0,0 +1,10 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+        package="me.phh.treble.overlay.xiaomi.redmi7onc"
+        android:versionCode="1"
+        android:versionName="1.0">
+        <overlay android:targetPackage="android"
+                android:requiredSystemPropertyName="ro.vendor.build.fingerprint"
+                android:requiredSystemPropertyValue="+*iaomi/onc_*"
```
TadiT7: `ro.vendor.build.fingerprint=xiaomi/onc_eea/onc:9/PKQ1.181021.001/V10.3.4.0.PFLEUXM:user/release-keys`
`ro.vendor.build.fingerprint=xiaomi/onc/onc:9/PKQ1.181021.001/V10.3.3.0.PFFINXM:user/release-keys`
`ro.vendor.build.fingerprint=xiaomi/onc/onc:9/PKQ1.181021.001/V10.3.4.0.PFLCNXM:user/release-keys`

`onc*` is needed to support all variants

phhusson:
Merged with @TadiT7's request in https://github.com/phhusson/vendor_hardware_overlay/commit/ec6ec7738d20f9eacbb9868c30803d40683acaee

PR #126, opened 2019-07-28
Add Overlay for Infinix S4 (X626)

dmurimi14:
- **Infinix S4 Overlay** serves majorly to fix the notch configuration and automatic brightness on treble custom Roms.
- **Infinix ZERO 6 Overlay** fixes automatic brightness on custom Roms especially treble GSIs while at the same time bringing fixes for the notch configuration including proper handling of WhatsApp statuses display.

### DIFF
--- a/Infinix/S4/res/values/notch.xml
+++ b/Infinix/S4/res/values/notch.xml
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+/*
+** Copyright 2017, The Android Open Source Project
+**
+** Licensed under the Apache License, Version 2.0 (the "License");
+** you may not use this file except in compliance with the License.
+** You may obtain a copy of the License at
+**
+**     http://www.apache.org/licenses/LICENSE-2.0
+**
+** Unless required by applicable law or agreed to in writing, software
+** distributed under the License is distributed on an "AS IS" BASIS,
+** WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+** See the License for the specific language governing permissions and
+** limitations under the License.
+*/
+-->
+<resources>
+    <!-- Height of the status bar in portrait -->
+    <dimen name="status_bar_height_portrait">50px</dimen>
+    <!-- Height of the status bar -->
+    <dimen name="status_bar_height">50px</dimen>
+    <!-- Height of the status bar in landscape -->
+    <dimen name="status_bar_height_landscape">28.0dip</dimen>
+</resources>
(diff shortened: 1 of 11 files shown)


Infinix/S4/res/values/notch.xml
```diff
@@ -0,0 +1,26 @@
...
+** WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+** See the License for the specific language governing permissions and
+** limitations under the License.
+*/
+-->
+<resources>
+    <!-- Height of the status bar in portrait -->
+    <dimen name="status_bar_height_portrait">50px</dimen>
+    <!-- Height of the status bar -->
+    <dimen name="status_bar_height">50px</dimen>
```
TadiT7: S4 doesn't need `status_bar_height`, it started with Pie. Tho overall, it's mergeable.

dmurimi14: I tried one GSI and there were issues with the status bar being pushed below the notch when using a higher dpi than the default one. This is what made me work on the overlay

dmurimi14: > S4 doesn't need `status_bar_height`, it started with Pie. Tho overall, it's mergeable.

I tried one GSI and there were issues with the status bar being pushed below the notch when using a higher dpi than the default one. This is what made me work on the overlay

phhusson: Thing is, even if I merge it, I'll remove the status_bar_height from pie branch anyway

dmurimi14: > Thing is, even if I merge it, I'll remove the status_bar_height from pie branch anyway

No problem sir, I understand

TadiT7: > > S4 doesn't need `status_bar_height`, it started with Pie. Tho overall, it's mergeable.
> 
> I tried one GSI and there were issues with the status bar being pushed below the notch when using a higher dpi than the default one. This is what made me work on the overlay

I meant just `status_bar_height`, not `status_bar_height_portrait`. No sane Pie ROM should use the former. BTW, if you meant that notch was bigger than the statusbar with "higher dpi" set, then it wasn't dpi what you changed, but dp display size in dev settings, which is basically reverse of dpi. The issue will be fixed either way.

dmurimi14: > > > S4 doesn't need `status_bar_height`, it started with Pie. Tho overall, it's mergeable.
> > 
> > 
> > I tried one GSI and there were issues with the status bar being pushed below the notch when using a higher dpi than the default one. This is what made me work on the overlay
> 
> I meant just `status_bar_height`, not `status_bar_height_portrait`. No sane Pie ROM should use the former. BTW, if you meant that notch was bigger than the statusbar with "higher dpi" set, then it wasn't dpi what you changed, but dp display size in dev settings, which is basically reverse of dpi. The issue will be fixed either way.

I get you know

PR #129, opened 2019-08-18
Add resources for Nubia Z17S (nx595j)

AndyCGYan:


AndyCGYan:
Closed - the way I see it, Z17S is not worth supporting, and mine has been sold already.

PR #130, opened 2019-08-19
B2N: overlay: Fix rounded_corner_radius

raysenlau:
Screen rounding set to 16dp would be a good fit


phhusson:
Uh you're really sure it works? 
As far as I know this property should be set in an overlay for SystemUI not framework-res

raysenlau:
> Uh you're really sure it works?
> As far as I know this property should be set in an overlay for SystemUI not framework-res

Yes, it works, I have tested it on PE.

raysenlau:
Similar to this commit：
https://github.com/phhusson/vendor_hardware_overlay/commit/52424fab69a148053484f8b46d108f0a390c3a45

TadiT7:
> Uh you're really sure it works?
> As far as I know this property should be set in an overlay for SystemUI not framework-res

Nah, there's `rounded_corner_radius` in framework-res, we are using that. What's SystemUI-exclusive is only the padding.

TadiT7:
Rebase on the current master branch please.

PR #131, opened 2019-08-22
Add support for Asus ZenFone Max (M2)

egor21312:
I have made overlay for this device to get properly working notch.


TadiT7:
Add `treble-overlay-asus-zenfonemaxm2` to overlay.mk

TadiT7:
Or whatever.

PR #134, opened 2019-09-22
Overlay for L38111/Lenovo Z6 Youth Edition/Lenovo K10 Note (kunlun2)

deanlancer:


Lenovo/L38111/res/values/config.xml
```diff
@@ -0,0 +1,159 @@
...
+        <item>275</item>
+        <item>481</item>
+        <item>618</item>
+        <item>756</item>
+        <item>989</item>
+        <item>1305</item>
+        <item>1511</item>
+        <item>2061</item>
+        <item>2748</item>
+        <item>3100</item>
```
phhusson: Auto brightness really works with this?

deanlancer: Yes, it does, actually. All the configs have been taken from the official rom.

phhusson: Official ROM can have brightness go up to 3100, but AOSP's max is 255

deanlancer: I see, I will change the brightness levels.

Lenovo/L38111/AndroidManifest.xml
```diff
@@ -0,0 +1,10 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+        package="me.phh.treble.overlay.lenovo.l38111"
```
TadiT7: Change to kunlun2, since `l38111` tells even less

overlay.mk
```diff
@@ -32,6 +32,7 @@ PRODUCT_PACKAGES += \
 	treble-overlay-huawei-WAS \
 	treble-overlay-huawei-telephony \
 	treble-overlay-lenovo-k5pro \
+	treble-overlay-lenovo-l38111 \
```
TadiT7: Same as above

TadiT7 (changes requested):

TadiT7:
Now it's mergeable, but I wish you could rename the directories first

TadiT7 (approved):

PR #135, opened 2019-10-29
Add overlay for Samsung A40

ponces:


Samsung/A40/res/values/brightness.xml
```diff
@@ -0,0 +1,596 @@
...
+        <item>303</item>
+        <item>319</item>
+        <item>335</item>
+        <item>351</item>
+        <item>367</item>
+        <item>383</item>
+        <item>399</item>
+        <item>415</item>
+        <item>431</item>
+        <item>447</item>
```
phhusson: This won't work properly, AOSP doesn't know anything but 0-255

At the moment, for this device, it's not possible to go higher than 255, though it's easy to add support for going up to 447 (it's in frameworks/base LightsService.java)

My recommendation is to truncate this array to 255

Samsung/A40/res/values/notch.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+<dimen name="status_bar_height_portrait">72px</dimen>
+<dimen name="status_bar_height_landscape">72px</dimen>
+<string name="config_mainBuiltInDisplayCutout"/>
```
phhusson: Well, Galaxy A40 *does* have display cutout.

Though i'm guessing waterdrop hurts pretty much noone, so we don't care, and it looks like Samsung didn't care as well, else this wouldn't be empty.

@TadiT7 what's you're recommendation there?

Samsung/A40/res/values/brightness.xml
ponces: Should the other values be also truncated to 255?
- config_autoBrightnessDisplayValuesNits
- config_screenBrightnessNits
- config_autoBrightnessLevels

ponces:
> I'm happy to see that my overlay generator is arlready put to good use :-)
> 
> Do you have a list of features you know you want to fix with this PR?
> Are they all fixed by this overlay, or are there other issues that might be fixed in an overlay that isn't there?

The purpose was to fix auto brightness on this device as the property didn't appear at all in the Display Settings page...

Samsung/A40/res/values/notch.xml
ponces: I'm lost here 😁

Samsung/A40/res/values/brightness.xml
phhusson: yup, all arrays must have the same size. (there might be some arrays that are size +1, or size -1, check their current size before truncating)

To be clear, when I say "truncate" I mean remove all the values above 255, not replace with 255

ponces: Yeah, I know. Thank you anyway :)
It should be fixed now.

Samsung/A40/res/values/notch.xml
ponces: This should be fixed with the latest commit (new values retrieved from framework-res__auto_generated_rro.apk)

ponces:
Please, check the latest commit

phhusson:
You've checked auto brightness works with this overlay?

ponces:
> You've checked auto brightness works with this overlay?

I have been testing it during this week and it doesn't seem to be working, even if the option is enabled in the Settings page (adaptive brightness)...
Is there any configuration or debug message that I can search for in a logcat file?

phhusson:
`adb logcat -b all |grep -i brightness -e backlight` on boot should give you some infos I think

diggit:
Hi,
I'd really like to have Adaptive brightness working on A40. Did you give up on it @ponces ?
[Here is gist with grepped log](https://gist.github.com/diggit/151fdf6116d4218beb0311e9e40f0ee4).

PR #136, opened 2019-11-01
Add support for Redmi Note 8 (ginkgo)

DarkJoker360:


phhusson:
Does it work? 
Sounds like auto brightness won't work

DarkJoker360:
I am going to test today in a while, I imported initial stuff for now. Will report back

PR #138, opened 2019-11-17
Add overlay for sony/kirin

immerblind:
overlay for xperia 10/kirin. battery usage stats fixed, also auto brightness seems to work...


phhusson:
Remove apns.xml

PR #139, opened 2019-11-19
Overlay for Lenovo Z5s(L78071)

ghost:
worked：
- auto brightness
- notch/status_bar
- brightness change


phhusson:
Looks good to me. @tadi777 ?
Should we remove values-land?

TadiT7:
> Looks good to me. @TadiT7 ?
> Should we remove values-land?

Here, yup, it's a Pie device, also the height should be changed to px

Lenovo/Z5s/res/values/notch.xml
```diff
@@ -17,9 +17,7 @@
 */
 -->
 <resources>
-    <dimen name="rounded_corner_radius">30dip</dimen>
-    <dimen name="status_bar_height">30.0dip</dimen>
-    <dimen name="status_bar_height_landscape">30.0dip</dimen>
-    <dimen name="status_bar_height_portrait">30.0dip</dimen>
+    <dimen name="rounded_corner_radius">30dp</dimen>
```
TadiT7: Is it a device with 480 dpi by default? Then that would translate to 90px instead, as rounded corner radius is physical.

pantanshin: Ye, thank you,i have modified

mrgreywater:
I don't think those build files are supposed to be in this PR, right?

```
build/aapt_mac 
build/build_mac.sh 
build/libc++.dylib 
build/signapk/libconscrypt_openjdk_jni.dylib 
build/signapk/signapk_mac.jar 
```

Lenovo/Z5s/res/values/notch.xml
```diff
@@ -0,0 +1,23 @@
...
+** Unless required by applicable law or agreed to in writing, software
+** distributed under the License is distributed on an "AS IS" BASIS,
+** WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+** See the License for the specific language governing permissions and
+** limitations under the License.
+*/
+-->
+<resources>
+    <dimen name="rounded_corner_radius">90px</dimen>
+    <dimen name="status_bar_height">87px</dimen>
```
mrgreywater: `status_bar_height_landscape` and `status_bar_height_portrait` seems to be missing

Lenovo/Z5s/res/values/notch.xml
```diff
@@ -0,0 +1,23 @@
...
+    <string translatable="false" name="config_mainBuiltInDisplayCutout">M -80,0 L -50,90 L 50,90 L 80,0 Z</string>
```
mrgreywater: I don't actually know how to interpret that path, but I find it curious that it only has coordinates divisible by 10, but the `status_bar_height` is not divisible by 10. Shouldn't the cutout height be equal to `status_bar_height`?

pantanshin:
> I don't think those build files are supposed to be in this PR, right?
> 
> ```
> build/aapt_mac 
> build/build_mac.sh 
> build/libc++.dylib 
> build/signapk/libconscrypt_openjdk_jni.dylib 
> build/signapk/signapk_mac.jar 
> ```

Maybe， 
Just thinking that other people using macOS can build

PR #149, opened 2019-12-05
Create config.xml for Asus Zenfone 5

emincangencer:


phhusson:
What does this fix?

emincangencer:
It's supposed to fix auto brigtness and some device features but with gapps it doesn't work properly. I am trying to understand why. After I fix the boot errors I will update it and create new request.

PR #151, opened 2019-12-06
Redmi7A: Disable rounded corners

YaAlex3:
its temporary, I don't know why roundness is too big tbh.


phhusson:
uh 106px indeed sounds very high, i'm more used to 26px
You can set it in treble app, so is there a point in removing it...?

YaAlex3:
i don't know honestly about normal values, but in miui framework was these values.

YaAlex3:
also is there any ways we can measure that?

YaAlex3:
any updates on this?
looks like i can't disable those in app..

PR #156, opened 2019-12-11
Add overlay for HuaweiEVA-AL10

jpourmalek:
it need to add this prop to vendor/build.prop
ro.hw.oemName=EVA-AL10

### DIFF
--- a/Huawei/hi3650/EVA/Android.mk
+++ b/Huawei/hi3650/EVA/Android.mk
@@ -1,0 +1,8 @@
+LOCAL_PATH := $(call my-dir)
+include $(CLEAR_VARS)
+LOCAL_MODULE_TAGS := optional
+LOCAL_PACKAGE_NAME := treble-overlay-huawei-EVA
+LOCAL_MODULE_PATH := $(TARGET_OUT)/overlay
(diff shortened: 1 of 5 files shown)


phhusson:
Well if it needs to change vendor, then i don't want that.
Use a property available in vendor without modification.

Le mer. 11 déc. 2019 à 23:56, jpourmalek <notifications@github.com> a
écrit :

barisdemirdelen:
Can we actually merge this? It works really well on my EVA-AL10 without any changes to /vendor

phhusson:
Do you know why @jpourmalek said they needed to add that to their vendor/build.prop...?

barisdemirdelen:
I don't know but my ro.hw.oemName is already EVA-AL10. I haven't changed it. Other huawei overlays also use ro.hw.oemName property.

if you like we can chane it to ro.vendor.build.fingerprint which starts with: HUAWEI/EVA-AL10

phhusson:
And what does this overlay fix for you?

barisdemirdelen:
Without is my battery life is around 1 hour and it gets very hot. I believe this overlay solves the issue

PR #157, opened 2019-12-15
Add Moto E5 (nora) Overlay

XavierWrz:
This enables adaptative brightness and battery stats


phhusson:
uh? Why did you close it?

PR #159, opened 2019-12-15
Add moto e5 (nora) overlay

XavierWrz:
This enables battery stats and adaptative brightness

### DIFF
--- a/Moto/E5/res/values/config.xml
+++ b/Moto/E5/res/values/config.xml
@@ -1,0 +1,121 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+	<bool name="config_enableAutoPowerModes">true</bool>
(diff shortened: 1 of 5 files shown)


Moto/E5/res/values/config.xml
```diff
@@ -0,0 +1,121 @@
...
+		<item>90</item>
+		<item>112</item>
+		<item>134</item>
+		<item>150</item>
+		<item>158</item>
+		<item>213</item>
+		<item>255</item>
+		<item>255</item>
+		<item>255</item>
+		<item>256</item>
```
phhusson: That sounds wrong, but if you tell me it works...

XavierWrz: i thought the same, but works

PR #163, opened 2019-12-22
WIP: Xiaomi Mi A3 preliminary support for auto brightness, power profile and notch…

lead4good:
…. Bluetooth still not working
@phhusson I am working on overlays for both this device and the Redmi Note 8. Both have the Snapdragon 665 chip and with both I experience the same issue with bluetooth audio. I have tried to use a bluetooth headset for both calling and audio playback. Apart from taking long to connect to the headset and connection interrupts, I was not able to playback audio via the bluetooth device. I was however able to make a call via the headset. Do you have an idea, what the problem could be?

You can find attached a logcat [bluetooth.log](https://github.com/phhusson/vendor_hardware_overlay/files/3992381/bluetooth.log) of trying to connect to a bluetooth speaker and playing back audio. There is a prominent error at the beginning regarding missing `android.permission.BLUETOOTH` permission. I do not know how to set them. And I also tried with different music apps and the browser, all playing back the audio via the phone.

The bluetooth settings I have found in the framework-res.apk regarding bluetooth were these:
```
    <integer name="config_bluetooth_idle_cur_ma">0</integer>
    <integer name="config_bluetooth_operating_voltage_mv">0</integer>
    <integer name="config_bluetooth_rx_cur_ma">0</integer>
    <integer name="config_bluetooth_tx_cur_ma">0</integer>
```
But they change nothing regarding the behaviour and don't seem to be correct either.

### DIFF
--- a/Xiaomi/MiA3/res/values/notch.xml
+++ b/Xiaomi/MiA3/res/values/notch.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+/*
+** Copyright 2017, The Android Open Source Project
+**
+** Licensed under the Apache License, Version 2.0 (the "License");
+** you may not use this file except in compliance with the License.
+** You may obtain a copy of the License at
+**
+**     http://www.apache.org/licenses/LICENSE-2.0
+**
+** Unless required by applicable law or agreed to in writing, software
+** distributed under the License is distributed on an "AS IS" BASIS,
+** WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+** See the License for the specific language governing permissions and
+** limitations under the License.
+*/
+-->
+<resources>
+    <!-- Height of the status bar in portrait -->
+    <dimen name="status_bar_height_portrait">35.0dip</dimen>
+    <!-- Height of the status bar in landscape -->
+    <dimen name="status_bar_height_landscape">28.0dip</dimen>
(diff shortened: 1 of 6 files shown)


phhusson:
I'd like to merge this without having bluetooth fixed yet (I doubt fixing bluetooth will be through overlay anyway).

For bluetooth, please open a new issue and/or respond on Moto G6 issue

lead4good:
Sure, go ahead!

Xiaomi/MiA3/res/values/notch.xml
```diff
@@ -0,0 +1,25 @@
...
+** Unless required by applicable law or agreed to in writing, software
+** distributed under the License is distributed on an "AS IS" BASIS,
+** WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+** See the License for the specific language governing permissions and
+** limitations under the License.
+*/
+-->
+<resources>
+    <!-- Height of the status bar in portrait -->
+    <dimen name="status_bar_height_portrait">35.0dip</dimen>
```
TadiT7: It should be an absolute px value

lead4good: Overlay has been tested on real device since about a week. Auto brightness works, status bar seems a bit to large and there is only space for 3 notifications even though foru notifications could very well fit before the notch covers them.

lead4good: How am I able to get the absolute px value?

TadiT7: > How am I able to get the absolute px value?

Preferably manually with some graphical mockups. Looks like the value should be around 53px.

PR #165, opened 2019-12-30
Redmi Note 8: support for auto brightness, power profile and notch

lead4good:


Xiaomi/RedmiNote8/res/values/config.xml
```diff
@@ -0,0 +1,199 @@
...
+        <item>905</item>
+        <item>1028</item>
+        <item>1152</item>
+        <item>1284</item>
+        <item>1418</item>
+        <item>1600</item>
+        <item>1760</item>
+        <item>1900</item>
+        <item>2000</item>
+        <item>2047</item>
```
phhusson: Aosp only accepts up to 255. Does auto brightness work for you?

PR #166, opened 2020-01-04
add Qin 2 Pro from DuoQin

PeterCxy:


phhusson:
I'll allow

wifi_tether_configure_ssid_default, wifi_localhotspot_configure_ssid_default and config_bluetooth_max_connected_audio_devices

However, could you explain to me why you want to set config_notificationsBatteryMediumARGB to green? medium battery level shouldn't be green

PeterCxy:
I'm pretty sure 0xFFFF0000 is red tho...

phhusson:
ah right sorry.
Still full red for medium is weird

PeterCxy:
nvm, that's probably just something I forgot to delete from official overlay. I don't think this LED is even RGB

PeterCxy:
removed

PR #167, opened 2020-01-13
Add resources for Samsung Galaxy A60 (a60q) / M40 (m40)

AndyCGYan:


phhusson:
The power_profile is obviously wrong, I'm removing it.

PR #168, opened 2020-01-15
Adds Samsung Galaxy A50

milankragujevic:


milankragujevic:
@phhusson I still get emails from people about this. Care to merge?

PR #172, opened 2020-01-28
Add overlay for Samsung A30

crazo7924:
Samsung A30 has many variants hence I used "a30*" in android:requiredSystemPropertyValue


crazo7924:
I think it needs removal of some values. tests.sh was warning about it

crazo7924:
Updated this request. Forgot to update overlay.mk

PR #175, opened 2020-02-08
Fix product name

GodcraftPlata:
So yeah, almost nobody uses GSIs on Hannah, so I think is better to use Rhannah product name


phhusson:
what? if you need both just do `+*hannah`

GodcraftPlata:
oh okay then

GodcraftPlata:
like this?

phhusson:
Yeah. Please confirm it works for you.

Le sam. 8 févr. 2020 à 23:30, GodcraftPlata <notifications@github.com> a
écrit :

GodcraftPlata:
Working

PR #176, opened 2020-02-19
Realme X2 Pro Vendor Overlay

mKenfenheuer:


phhusson:
Please no space in folder name

Realme/X2 Pro/AndroidManifest.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8" standalone="no"?><manifest xmlns:android="http://schemas.android.com/apk/res/android"  package="me.phh.treble.overlay.realme.x2pro" android:requiredSystemPropertyValue="+realme/RMX1931EEA*">
+    <overlay android:isStatic="true" android:priority="108" android:targetPackage="android"/>
```
phhusson: it is missing device-specific match

mKenfenheuer: Thats right. unfortunately the vendor build.prop does not provide any values to match against.

Add to /vendor/build.prop
`# Patch to add device specific match for treble overlays by mkenfenheuer
ro.vendor.build.overlay.deviceid=RMX1931`

I've added a value called ro.vendor.build.overlay.deviceid to get around that, but i missed to commit the changes to the manifest. 

There is also a pull request on my repo, which will enable the 90hz display, but i'd prefer your setting inside the treble settings app as you can choose between 90 and 60hz instead of forcing the users to 90hz.

I'll update the repo asap.

mKenfenheuer: Works (confirmed by me): 
- Always on display

'Should' be working:
- Enables DT2W option but enabling does not seem to have any effect to me.
- VOLTE - i cannot test as my carrier does not offer this anymore
- more reliable network connection (i have noticed far less lte issues, but this is my perception)

MISC:
Imported the settings from realme vendor (i don't know if they are nescessary)

mKenfenheuer:
If i just build the overlay, the refresh rate works. But test are complaining that the value config_defaultPeakRefreshRate is not defined ?

phhusson:
Add it to knownKeys

phhusson:
To detect the hardware, the best I found is:
```
if [ -b  /dev/block/bootdevice/by-name/oppodycnvbk ];then
    for devname in RMX1981 /* insert other oppo codenames here */;do
        if grep -q $devname  /dev/block/bootdevice/by-name/oppodycnvbk;then
            setprop ro.build.overlay.deviceid $devname
       fi
    done
fi
```

phhusson:
Done and commited https://github.com/phhusson/device_phh_treble/commit/62f63e975d33a253941074286b267bf26f48e28a#diff-d07238f52cec7f0583ea064336a0ddd2R156
You can use ro.build.overlay.deviceid  in your overlay for matching

phhusson:
thanks

PR #178, opened 2020-02-26
vendor_hardware_overlay for Moto Z2 Force 

thaiminhan:
Base on Boot and Vendor from Lineageos 17.1 Erfan https://mirrors.lolinet.com/firmware/moto/nash/custom/lineage/lineage-17.1-20200208-UNOFFICIAL-nash.zip


thaiminhan:
fix non-existing attribute

PR #180, opened 2020-02-27
Add Umidigi Power overlay

DeadSquirrel01:


phhusson:
It's tested and it works?

You don't need to add automatic brightness settings in addition to that?

DeadSquirrel01:
Yes, I tested it. Adaptive brightness already works

phhusson:
Ok thanks.
FWIW you didn't add the entry in overlay.mk at the right place (it's lexical order)
CI should have said that, but looks like github didn't run the CI. :man_shrugging:

PR #186, opened 2020-03-11
Lenovoz5s

bssbrubs:
Just a overlay for Lenovo Z5s, working well, we tested on brazilian group on telegram, t.me/lenovoz5sbr


mrgreywater:
There's already PR #139

PR #187, opened 2020-03-15
Remove redundant config_dozeComponent property

Dunedan:
The `config_dozeComponent` property is already [set via
device_phh_treble][1] and all overlays overwriting it just did so with
its default value. As this doesn't add any value, this commit removes
the definition of `config_dozeComponent` from all overlays which set it.
It also removes `config_dozeComponent` from `knownKeys` with the
intention that somebody wanting to overwrite it in future should justify
why it should get added back to `knownKeys` again.

[1]: https://github.com/phhusson/device_phh_treble/blob/04583fa95a6c58d25dda7644002f9da05eee3ab2/overlay/frameworks/base/core/res/res/values/config.xml#L32


phhusson:
Please even add it to blacklist, running tests on my computer might add it again by accident

Dunedan:
Done. :slightly_smiling_face:

PR #190, opened 2020-03-15
Remove some redundant properties

Dunedan:
This PR removes some properties which were either set to the default of AOSP or not used at all and blacklists them. Further details are in the respective commit messages.


phhusson:
Well there is a use for pinnerHomeApp (and pinnerCameraApp alongside), i'm suprised noone uses it.
It is used to pin the launcher and the camera app into memory, so that it is faster to start.

This is to be defined based on RAM available on the device.
In my dreams, I'd have a RAM-related overlay (because it is not related to the hardware, just how much memory it has), like hardware_overlay/RAM/2G; 3G: 4G; 6G, with settings for those respective sizes, and I'd enable those via a property from rw-system.sh

Dunedan:
Sounds good. Shall I just revert the `config_pinnerHomeApp` commit and you merge the other two?

phhusson:
I'd rather you push -f without that commit

Dunedan:
Here you go.

PR #192, opened 2020-03-16
Add overlay for Realme C1

Tuan9304:
Initial commit

### DIFF
--- a/Realme/C1/Android.mk
+++ b/Realme/C1/Android.mk
@@ -1,0 +1,8 @@
+LOCAL_PATH := $(call my-dir)
+include $(CLEAR_VARS)
+LOCAL_MODULE_TAGS := optional
+LOCAL_PACKAGE_NAME := treble-overlay-realme-c1
+LOCAL_MODULE_PATH := $(TARGET_OUT)/overlay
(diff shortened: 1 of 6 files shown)


phhusson:
Have you actually tested it?
In my experience, `ro.vendor.build.fingerprint` doesn't work on Oppo devices

Tuan9304:
Oh sorry I don't know that. Do you have any idea? Can I use `ro.hw.oemName`?

Tuan9304:
It works fine now

Tuan9304:
My pull request to fix brightness: https://github.com/phhusson/device_phh_treble/pull/165

phhusson:
Auto brightness values are pretty fishy. Have you actually tested it?

Tuan9304:
Yes, it work too. There are `<"config_autoBrightness...">` & `<"config_tenbit_autoBrightness...">` & `elevenbit` and `twelvebit`. 
I tried `config_autoBrightness` with max value is 255 but it not work correctly (brightness so dark when not maximized) so I tried `elevenbit` with the same max brightness value with max_brightness file and it works well.

Tuan9304:
It seems everything is fine. I can't test voLTE because my carrier doesn't support it.

PR #197, opened 2020-03-25
Change config.xml to support VoLTE

khrj:
Redmi 7A Supports VoLTE with https://github.com/KhushrajRathod/VoLTE-Fix


phhusson:
What does this overlay actually fix?

There is nothing device-specific in there.

KhushrajRathod:
It's not possible to use VoLTE without using this overlay. I used the Ims-q apk and android.hardware.telephony.ims.xml along with this overlay change to enable VoLTE

phhusson:
That's weird, no other device require it
All devices go just fine with https://github.com/phhusson/vendor_hardware_overlay/blob/master/CAF-IMS/res/values/config.xml

Perhaps the issue is that those flags are in this overlay in the first place? I don't know who put them there and why

Also, `config_wifi_dual_band_support` has nothing to do with VoLTE.

KhushrajRathod:
Will remove flags completely and check VoLTE function, then revert back to you

KhushrajRathod:
Nope, completely removing the flags disables VoLTE. The flags are required for VoLTE to function.

KhushrajRathod:
I've changed the config_wifi_dual_band_support flag

KhushrajRathod:
Please merge

phhusson:
Did the equivalent in f4e0df4

Thanks for bringing up this issue.

PR #198, opened 2020-03-25
Enable custom location provider support

jumper047:
Seems like i figure out why microg don't work on my redmi:)


phhusson:
I'm wondering if I shouldn't just remove that overlay, it seems totally fucked up...

jumper047:
Actually  it works fine for a lot of people. I don't see any complains in phone's telegram group, xda or other local forums. But, anyway, this is your project, and i am not an android developer. Just a guy who can read article from your wiki and follow it's instructions. So you can delete it anytime - it works for me, and for me it is enough. Just wanted to give back something to community.

jumper047:
I'll try to check it and remove unnecessary params later.

PR #203, opened 2020-04-03
Enable dt2w for begonia / Redmi Note 8 Pro

TTTT555:


phhusson:
You've tested it?

TTTT555:
> You've tested it?

A switch in the menu appeared, but the function does not work, additional edits are needed (the function worked in the Zaoda firmware).

phhusson:
K so it doesn't work.

Have you tried dt2w setting in treble settings?

TTTT555:
> K so it doesn't work. Have you tried dt2w setting in treble settings?

I don't see this setting in the menu (LineageOS 17.1)

TTTT555:
Not needed with Xiaomi features in PHH Treble App

PR #206, opened 2020-04-10
Updated overlay for Galaxy A30

crazo7924:
This removes two lines in `arrays.xml`. Both causes the system to throw an error and not let the system boot.


phhusson:
thanks

PR #207, opened 2020-04-12
Remove device specific IMS settings

Dunedan:
This removes all device specific IMS settings, which are already set by the *-IMS overlays.

As a result `config_ims_package` could be added to the blacklist of keys, if it wouldn't be used in the *-IMS overlays.


phhusson:
Thanks

PR #210, opened 2020-05-01
Add overlay for Samsung A40

ponces:
A few months ago, I made a pull request just like this one for A40 but brightness was not functioning at all. But now it is! It has been tested and I'm even currently using this overlay on my daily driver.


phhusson:
Looks good to me.

Can you spend a bit of time to cleanup values to merge those that goes together?

Something like https://github.com/phhusson/vendor_hardware_overlay/pull/209

ponces:
Thank you for the quick reply!
Those are already a list of filtered values.
If you compare them with the entries used by A30 (a very similar device) you won't find any difference, only in the values themselves.

phhusson:
I'm speaking of the fact of having XML file names actually describe what it
does like having brightness.xml for brightness stuff

Le ven. 1 mai 2020 à 18:35, Alberto Ponces <notifications@github.com> a
écrit :

ponces:
Thank you for the tip.
Is this ok now?

phhusson:
Yup perfect

PR #215, opened 2020-05-14
A30: respect corners

crazo7924:
Density independent pixels unit is much better.

Tested on self-built LineageOS ROM


phhusson:
Uh sorry what? If you take dp-s, that means the padding changes with user-set density?
I don't understand why you would want that...?

Like I like things small, so Iower dpi, I'll have smaller corners? But then with lower padding, icons will hide behind the real screen corner?

crazo7924:
I'm closing this for now

phhusson:
But please come back later, I'm sure you did this for a reason that needs to be fixed :-)

crazo7924:
@phhusson I think I need to overlay SystemUI and not the core package

PR #216, opened 2020-05-22
Overlay Galaxy A10

Jazz256:
Hello I have added the overlay for the Samsung Galaxy A10, and I would like to request that it be added to your repository. Thank you.


phhusson:
Remove fail file.
Have you tested? Which features does it fix?

PR #221, opened 2020-06-08
Add overlay for Vsmart Live

aljjo:
Tested battery stats, brightness. Remeasured corner.

What does this do?
```
    <string-array name="config_ephemeralResolverPackage">
        <item>com.google.android.gms</item>
    </string-array>
```
I still don't understand after reading it's description, so it isn't included.


aljjo:
Tested headphone jack.

PR #223, opened 2020-06-14
[Nokia 8.1/X7] Add fingerprint gestures, fix statusbar content padding in landscape mode, stock-like statusbar icons behavior

TechnoStone:


phhusson:
Thanks. Can you be more precise for "Stock-like statusbar icons behaviour"?

TechnoStone:
It means that margins are same as on stock and there are placing: clock + 2/1+dot notifications on the left side of statusbar and 5/4+dot icons on the right side of statusbar due to large notch

phhusson:
ok thanks. status_bar_landscape in px is weird, but whatever

TechnoStone:
We was like on this picture
Untitled-1

PR #241, opened 2020-08-06
Update strings.xml for S10q hole punch spacing

Sn0whax:
<string name="config_mainBuiltInDisplayCutout">M 0,0 H -68 V 42.57142857142857 H 0 V 0 Z @dp @right</string>


phhusson:
hum? Can you maybe show a photo before/after? or better a screenshot with config_fillMainBuiltInDisplayCutout = true

PR #257, opened 2020-09-21
Add overlay for Xiaomi Redmi Note 9 (merlin)

SandaruLJ:
- Fixes auto brightness
- Fixes headphone jack


phhusson:
This has been tested on my AOSP GSI?

SandaruLJ:
> This has been tested on my AOSP GSI?

Yes. Tested on your latest AOSP and several other GSIs.

phhusson:
I don't understand how it can work with brightness values going up to 2047, when AOSP only accepts up to 255...

SandaruLJ:
First I had to set `persist.sys.qcom-brightness` to 2047. Only then the brightness slider works as expected. Both Redmi Note 8 Pro and Redmi Note 9 has brightness values up to 2047.

Akitakek:
Either way I'd like to see progress on this, the brightness slider keeps making brightness values go up and down and up as you slide it across, it's really annoying.

PR #270, opened 2020-10-03
Add overlay for Vsmart Live (V620A)

trinhloi:
Fix AOD,Brightness,Fingerprint


trinhloi:
> Fix AOD,Brightness,Fingerprint

PR #276, opened 2020-10-07
[daisy] fix wrong statusbar height

eremitein:
broken after 542ff298281f7e926f3188aabd430155e87560ce


phhusson:
that's a bit wtf, and it's even more wtf it has been there for so long but well

PR #304, opened 2020-12-16
add Spreadtrum IMS overlay for R

PeterCxy:
Changes in Treble App also required


phhusson:
K the tests passed except for shit I pushed before

PR #313, opened 2021-01-08
add overlay moto e6 plus

LichtEstarossx:


Moto/E6Plus/AndroidManifest.xml
```diff
@@ -0,0 +1,11 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+        package="me.phh.treble.overlay.moto.e6plus"
+        android:versionCode="1"
+        android:versionName="1.0">
+
+        <overlay android:targetPackage="android"
+                android:requiredSystemPropertyName="ro.vendor.build.fingerprint"
+                android:requiredSystemPropertyValue="+*pokerp"
```
phhusson, line 8: Is it tested? I'm surprised this work

LichtEstarossx: seriously? go then I think the users who used it deceived me 😭😭

phhusson:
Why did you close it?
You no longer think it works?

LichtEstarossx:
For now I will try to do it again, but this time I will not have to rely too much on my few testers, I will do my best to make it work correctly, I am sorry for the inconvenience it has caused

PR #325, opened 2021-02-26
Update knownKeys

Hirbod-JORFlint:
config_enable_emergency_call_while_sim_locked >> Enable emergency call when sim is locked or puk locked(note: carrier dependent)
config_device_respects_hold_carrier_config >> controls whether ims call holding is possible or not
config_use_voip_mode_for_ims >> whether to use voip audio for ims calls

##below configs target telephony service

config_carrier_settings_enable >> If set to true, will display Carrier settings
config_carrier_volte_provisioning_required >> specify whether provisioning is required for VOLTE
dtmf_type_enabled >> enable dtmf tone type when set to true
allow_emergency_numbers_in_call_log >> controls whether to display emergency calls in call logs or not(default:false)
sim_network_unlock_allow_dismiss >> whether the Phone app should provide a "Dismiss" button on the SIM network unlock screen


phhusson:
Duplicate of 324

PR #329, opened 2021-03-04
Add overlay for Samsung Galaxy J6 2018

FraSharp:


FraSharp:
checked with xmlstarlet, built and tested in "https://github.com/phhusson/treble_experimentations/releases/download/v301/system-roar-arm64-ab-vndklite-vanilla.img.xz"

PR #335, opened 2021-03-24
(Galaxy A51) add overlay

KatoTempest:


phhusson:
Looks good to me, except that you define config_dynamic_automatic_brightness_available which doesn't exist

KatoTempest:
I had forgotten, I just deleted that line, although I noticed that it has a conflict with the backlight that is set in treble app, it must be set to 0 for the brightness to work, if it does not stay the same

PR #347, opened 2021-05-26
Add overlay for Xiaomi Redmi 9 (lancelot/galahad)

SebastianApel:


phhusson:
Is this tested?

SebastianApel:
Yes, it is tested.

The only glitch I observed is that the backlight stayed on at a very low level (only observable during nighttime), even when the phone is in sleep mode (and screen is blank). Do you have a suggestion how to fix that?

SebastianApel:
I've tested the overlay now for a couple of days, it works without problems.  The "glitches" I was experiencing was not due to the overlay, but to the fact that "persist.sys.qcom-brightness" was not set correctly.

Root cause (as far as I can determine) is as follows:
1. rw-system.sh has code for Redmi/galahad and sets qcom-brightness to the correct value (2047)
2. BUT: phh-prop-handler.sh overwrites this with -1, because "persist.sys.phh.backlight.scale" is set to 0
3. As a result, brightness does not work correctly

This means that the code in phh-prop-handler.sh conflicts with code in rw-system.sh. Are you aware of that? Do you want me to open an issue for that?

Regarding this PR: It's tested and it works.

SebastianApel:
@phhusson: Is there anything else that would need to done from your perspective before this PR can be merged?

phhusson:
Yes what you describe is known. Setting qcom-brightness from rw-system.sh is no longer supported, and you're supposed to make a treble_preset to include this setting, see https://github.com/phhusson/treble_presets/

SebastianApel:
OK, thank you merging and for the feedback! I have opened https://github.com/phhusson/treble_experimentations/issues/1933 and https://github.com/phhusson/treble_experimentations/issues/1932 for the follow-up on this - so this PR can be closed.

PR #348, opened 2021-05-26
Update AndroidManifest.xml

kevios12:
Overlay issue: 

from: android:requiredSystemPropertyValue="+*amsung/a40*"

to: android:requiredSystemPropertyValue="+*samsung/a40*"

greets


phhusson:
It's * so it matches so that's not an issue

PR #371, opened 2021-06-15
Add overlay for Samsung Galaxy M11

arijit192:
Adding overlay for Samsung Galaxy M11 (SM-M115F)


phhusson:
Looks good to me. Have you tested it?

arijit192:
I could not transfer the files to /system/product/overlay because the root is read-only. I tried to make /dev/block/dm-0 readwrite by blockdev but no luck. I also tried to modify system.img by mounting it to my PC but still no luck. Can you please suggest some workarounds?

phhusson:
Did you flash vndklite variant?

Le sam. 19 juin 2021 à 11:17, Arijit Roy ***@***.***> a
écrit :

arijit192:
No. This one system-roar-arm32_binder64-ab-gapps.img.xz

phhusson:
Well, with the nvdklite variant it'll be easier to mount / rw

Le dim. 20 juin 2021 à 03:40, Arijit Roy ***@***.***> a
écrit :

arijit192:
Okay thanks. I will test it ASAP and let you know

arijit192:
I have tested the vendor overlay. As per my testing, some things are not working like WiFi calling, VoLTE, MTP and no sound of voice when on audio call. Other than that everything is working fine.

phhusson:
Ok, but what does the overlay fix?

20 juin 2021 19:04:24 Arijit Roy ***@***.***>:

arijit192:
After observing well I noticed very little changes like the auto brightness is working a little better and notch management seems to be better. Apart from that I have not seen any huge difference between overlay and without overlay. Without overlay the device is working just fine except the above things that I mentioned in comment are broken.

PR #382, opened 2021-07-31
Add overlay for Xiaomi Mi 11 Lite 5G

AndroPlus-org:
Tested on Android 11 GSI.

By the way, Mi 11 Lite 5G for Japan (Xiaomi/renoir_jp) needs 4 apks in /system/app to use NFC Type-F features.
Is there any good way to integrate them into GSI but don't waste other device users' storage? They took about 12MB...
https://github.com/FlokoROM-GSI/vendor_hardware_overlay/tree/pie/FeliCa


phhusson:
Nope, at the moment I have no way to integrate vendor APK, though my current idea is to have a overlay (overlayfs kernel support is required since Android 10 vendors)

PR #391, opened 2021-09-18
support umidigi f1 and f1 play overlay

shadowofleaf96:
fix notch and statusbar height and other things


phhusson:
You need to add `treble-overlay-umidigi-f1` in overlay.mk

PR #394, opened 2021-10-05
Add Infinix Note 8 overlay

ryenyuku:


ryenyuku:
I removed the `config_showNavigationBar` key because it's already defined in the vendor's overlay on /vendor

PR #395, opened 2021-10-07
Changes and fixes for the galaxy a51 overlay

KatoTempest:


Samsung/A51/AndroidManifest.xml
```diff
@@ -4,7 +4,7 @@
         android:versionName="1.0">
         <overlay android:targetPackage="android"
         	android:requiredSystemPropertyName="ro.vendor.build.fingerprint"
-        	android:requiredSystemPropertyValue="+*samsung/a51*"
+        	android:requiredSystemPropertyValue="+*samsung/a51nsxx*"
```
phhusson: why?

Samsung/A51/res/xml/color_extraction.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
+<colorextraction>
```
phhusson: What does a palette have to do with hardware support?

Samsung/A51/res/values/config.xml
```diff
@@ -1,633 +1,98 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-	    <string-array name="config_tether_usb_regexs">
-        <item>rndis0</item>
+    <integer name="config_screenBrightnessSettingMaximum">255</integer>
+	<string-array name="config_tether_wifi_regexs">
```
phhusson: Why? Was wifi tethering previously broken?

Samsung/A51/res/values/config.xml
```diff
@@ -1,633 +1,98 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-	    <string-array name="config_tether_usb_regexs">
```
phhusson, line 3: Why? Was usb tethering previously broken?

Samsung/A51/res/values/config.xml
```diff
@@ -1,633 +1,98 @@
...
-        <item>rndis0</item>
+    <integer name="config_screenBrightnessSettingMaximum">255</integer>
```
phhusson: This value was already there...? Why move it away? It makes re-reading pretty hard

phhusson (commented):
I have absolutely no idea what the actual changes are, nor why, nor why you say it's minor

KatoTempest:
There were a few things broken regarding the usb, the change from a51 to a51nsxx is in case the overlay was applied on the a51 5g, also the fact that it had changed places some lines was because it was testing that the brightness It would work correctly since according to several users the brightness was either low or it did not work in automatic mode
although thinking about it, I think they were not just minor changes

PR #396, opened 2021-10-09
Add resources for OnePlus Nord N200 (OnePlusN200)

AndyCGYan:


AndyCGYan:
The empty `<application>` tag is there so that the overlay compiled by `build.sh` can be properly parsed - I've notified you about this on TG before.

OnePlus/N200/res/xml/power_profile.xml
```diff
@@ -0,0 +1,109 @@
...
+    <array name="wifi.controller.tx_levels" />
+    <item name="wifi.controller.voltage">3300</item>
+    <array name="wifi.batchedscan">
+        <value>.0002</value>
+        <value>.002</value>
+        <value>.02</value>
+        <value>.2</value>
+        <value>2</value>
+    </array>
+    <allow-in-power-save package="com.google.android.gms" />
```
phhusson, line 106: not that i care, but I'm pretty sure this shouldn't be there

AndyCGYan: Taken from stock as-is ofc

PR #401, opened 2021-10-28
InfinixNote8: Add SoftAP support

ryenyuku:
https://source.android.com/devices/tech/connect/wifi-softap-tethering
_Note: Since the key isn't defined in "knownKeys" file, I've added it for your pleasure_


phhusson:
Commit message is wrong?
You don't add support for SoftAP (which is already supported) but merely for SoftAP ACS?

ryenyuku:
Oh right, LOL

PR #406, opened 2021-10-31
Add MAC randomization supported on BV9500Plus

Vgdn1942:


phhusson:
Is it testede?

Vgdn1942:
> Is it testede?

Yes.

PR #414, opened 2021-11-15
BV9500Plus: Back brightness values to stock; Add BV9800Pro overlay

Vgdn1942:


phhusson:
Why?

Vgdn1942:
> Why?

Now it is too dark outside and in other bright places.

phhusson:
And this commit fixes it? It really shouldn't

Vgdn1942:
> And this commit fixes it? It really shouldn't

Yes. This returns the stock brightness values.

PR #415, opened 2021-11-15
Realme 2 and C1 Overlay.

custosoft:
I hope that i will push it. thanks


custosoft:
Please Add it in you git

PR #422, opened 2021-12-03
Add overlay for Infinix Note 10 Pro (X695C)

ramabondanp:


ramabondanp:
done @phhusson what am i missing or wrong doing?

PR #423, opened 2021-12-04
Add overlay for Sony pdx213 (Xperia 10 III)

Aarqw12:
Treble overlay for Sony pdx206 (Xperia 5 II) is also fully compatible for pdx213 and xperiadev overlay is pretty same for both devices.


phhusson:

PR #424, opened 2021-12-07
Add resources for OnePlus 8 (OnePlus8/TMO)

AndyCGYan:


OnePlus/OP8/AndroidManifest.xml
```diff
@@ -0,0 +1,10 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+        package="me.phh.treble.overlay.oneplus.op8"
+        android:versionCode="1"
+        android:versionName="1.0">
+        <overlay android:targetPackage="android"
+                android:requiredSystemPropertyName="ro.product.odm.device"
```
phhusson, line 6: Uh? That works? I seem to recall I recoded my own "fnmatch" that only does the `*`

PR #428, opened 2021-12-18
add moto e6plus

KatoTempest:
the overlay has slight changes in brightness and other values, the value assigned to vibration was set manually because the stock value was too low

### DIFF
--- a/Moto/E6Plus/Android.mk
+++ b/Moto/E6Plus/Android.mk
@@ -1,0 +1,8 @@
+LOCAL_PATH := $(call my-dir)
+include $(CLEAR_VARS)
+LOCAL_MODULE_TAGS := optional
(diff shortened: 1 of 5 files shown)


phhusson:
Good for me, you just need to pick a new priority, you can take 498

KatoTempest:
I think I will try to reduce the manual brightness more, at night it bothers the view

KatoTempest:
I had forgotten that I made the pull request, update my repository in case it has any conflict with the original branch (in case the overlay is accepted)

phhusson:
please rebase your branch on top on phhusson:pie rather than merge

KatoTempest:
I think I did the rebase correctly, it took a while due to the malfunction of my internet, although I don't see much difference in the repository

phhusson:

PR #430, opened 2021-12-21
Add overlay for Realme Q3 5G (RMX3161)

Cloverk1t:
Fix punch hole and battery capacity detection


phhusson:
wrong ordering: q is before x

Cloverk1t:
> wrong ordering: q is before x

Okay, it's fixed

PR #431, opened 2021-12-22
Samsung A71 overlay

adamfirari:


adamfirari:
Samsung a71 overlay

phhusson:
Wow that's a very big change. What happened?

adamfirari:
> 

First commit from overly i got from system
Second one  is from overly i got from  product

PR #436, opened 2022-01-05
Add overlay for exynos2100 devices

haridhayal11:
Auto Brightness works (Tested on S21 and hopefully on S21+ and S21Ultra)
Will update after testing on S21+ and S21Ultra.


haridhayal11:
Works on S21+ and S21Ultra too

PR #439, opened 2022-01-08
Add Motorola One Fusion Overlay

dpkg-i-foo-deb:
Adaptive brightness now works on this device, I took all the values from the stock framework-res.apk in the system partition, the status bar fits perfectly everywhere except on the lock screen, no idea why


phhusson:
```
Fatal: ./Moto/OneFusion: Overlay ./Moto/OneFusion is defining config_swipe_up_gesture_setting_available which is forbidden
Fatal: ./Moto/OneFusion: Overlay ./Moto/OneFusion is defining status_bar_height which is forbidden
Fatal: ./Moto/OneFusion/res/values/config.xml: defines a non-existing attribute status_bar_content_number_size
Fatal: ./Moto/OneFusion/res/values/config.xml: defines a non-existing attribute status_bar_edge_ignore
Fatal: ./Moto/OneFusion/res/values/config.xml: defines a non-existing attribute status_bar_icon_size
Fatal: ./Moto/OneFusion/res/values/config.xml: defines a non-existing attribute status_bar_padding_top_portrait
Fatal: ./Moto/OneFusion/res/values/config.xml: defines a non-existing attribute status_bar_system_icon_intrinsic_size
Fatal: ./Moto/OneFusion/res/values/config.xml: defines a non-existing attribute status_bar_system_icon_size
Fatal: ./Moto/OneFusion/res/values/config.xml: defines a non-existing attribute config_fingerprintSupportsGestures
```

dpkg-i-foo-deb:
Damn, I almost thought I wasn't going to be able to do it...

After removing the stuff causing the errors there is no difference in the ugly lock screen, everything else works just fine tho :D

dpkg-i-foo-deb:
This is the file I grabbed from the phone, the one that breaks the freaking lock screen T~T

file:///media/dpkg/Usen-GNU-Linux/Apps/Android-Tools/SystemUI__auto_generated_rro_vendor.apk

dpkg-i-foo-deb:
I had to compress it because github went mad, this is the apk causing the lock screen mess
[SystemUI__auto_generated_rro_vendor.zip](https://github.com/phhusson/vendor_hardware_overlay/files/7833477/SystemUI__auto_generated_rro_vendor.zip)

dpkg-i-foo-deb:
The file I sent contains this value     `<dimen` `name="global_actions_top_padding">506.0px</dimen>`

When set to 0 and pushed back in the problem is solved. However, I don't know how to apply that to my own overlay as it doesn't work if I include it manually

PR #442, opened 2022-01-12
Add Volte for Huawei device hi6250

Iceows:
Here is a patch which allows to support in Android 11 the volte via the HuaweiIMS ims. You also need to modify Treble App


phhusson:
Add `treble-overlay-telephony-hw-ims` to overlay.mk

Iceows:
Add overlay treble-overlay-telephony-hw-ims  to overlay.mk

PR #446, opened 2022-01-20
Update A51 overlays

Yilliee:
* Enable AoD and dt2w options by default on a51.
* Add a systemui overlay to align status bar icons with the punch hole cutout vertically.


phhusson:
Sounds wrong to me. config_supportDoubleTapWake assumes sensors HAL provide a dt2w sensor, which it doesn't, which is way i have a dt2w option in treble settings

PR #451, opened 2022-01-28
Update config.xml

wmdhs12138:
Minor fix for brightness


wmdhs12138 (commented):
1 is too dark, 2 will be fine.

PR #454, opened 2022-02-01
Add overlay for Galaxy A52

nicovanoo:


nicovanoo:
Tested on PE12 gsi, automatic brightness fixed.

PR #464, opened 2022-02-17
Device overlay for the Oppo A54

Johx22:


ponces:
@Johx22 don't ask me again for help!
I provided you an overlay apk to test and clearly you don't even know what you're doing and didn't even tested it... so if you have issues using this (if it will be merged at all) don't bother me to fix them

ponces:
I'm not even bothered to ask you for credits on this.
Anyway I don't want credits on stuff I don't support or even know if it's working

PR #471, opened 2022-02-21
Create sensors.xml for s9 series snapdragon

marsyeen:


marsyeen:
This attempts to fix the vibration strenght.

marsyeen:
This seems not to have fixed the issue.

PR #474, opened 2022-02-24
adding overlay files for the Moto G100 device

imqipan:


phhusson:
you forgot to add it to `overlay.mk`

PR #479, opened 2022-02-26
Add resources for Vivo Y20

Johx22:


Johx22:
Small correction: This applies to the vivo Y20 2021 model

PR #495, opened 2022-03-21
Added initial moto G31 support

polhdez:


polhdez:
I have added this device to the wiki and tested this overlay for half a day and everything seems to work flawlessly :)

PR #508, opened 2022-04-04
Made changes to A51 & M31s overlays

rufus582:
A51: Fix battery capacity (was set to 1000, I set it to 4000. Hope the battery usage data will work after this change)
M31s: Increase AOD brightness from 10 to 17

### DIFF
--- a/Samsung/A51/res/xml/power_profile.xml
+++ b/Samsung/A51/res/xml/power_profile.xml
@@ -35,7 +35,7 @@
     <array name="memory.bandwidths">
         <value>22.7</value>
     </array>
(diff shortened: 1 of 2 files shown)


phhusson:
> A51: Fix battery capacity (was set to 1000, I set it to 4000. Hope the battery usage data will work after this change)

It's very likely that if battery capacity was wrong, everything else is wrong as well...

rufus582:
Makes sense😂. I'm trying to get stock overlay file from A51, so then I can copy the power_profile.xml from that then🤔. I'll make another PR if I get that...

PR #523, opened 2022-04-30
Add resources for Samsung A53

Johx22:
Tested and verified by @dollscythe


Samsung/A53/res/values/configs.xml
```diff
@@ -0,0 +1,512 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <integer-array name="config_screenBrightnessBacklight">
```
phhusson, line 3: Sounds useless to me...? This array doesn't fix anything, does it?

Johx22: Tbh yes it doesn't.. It was there in arrays.xml so I just added them.

PR #531, opened 2022-05-05
Add Samsung Galaxy S22 (r0s) Overlay

jgudec:


phhusson:
Did you actually test it?

I'm disabling adaptive brightness on my S22 because it keeps blinding me in the dark...

jgudec:
> Did you actually test it?

I did test it in the following conditions:
1) Within a closed room, where there was pretty much no light ( I deemed this "okay" as it burned my eyes the same as stock)
2) Normally-lit living room
3) Under direct sunlight

To be fair, I hoped someone else would give it a try, provide feedback and/or possibly fine-tune it so an acceptable setting would be reached (instead of my clearly subjective one).

PR #536, opened 2022-05-10
Add overlay for Xiaomi 11T Pro

alexandreapis:


Xiaomi/11TPro/AndroidManifest.xml
```diff
@@ -0,0 +1,10 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+        package="me.phh.treble.overlay.xiaomi.11tpro"
+        android:versionCode="1"
+        android:versionName="1.0">
+        <overlay android:targetPackage="android"
+                android:requiredSystemPropertyName="ro.product.vendor.device"
```
diogotcorreia, line 6: Shouldn't this use the fringerprint instead? For me, `getprop ro.product.vendor.device` only prints out `vili`, which doesn't match this.
```suggestion
                android:requiredSystemPropertyName="ro.vendor.build.fingerprint"
```

diogotcorreia, line 6: Shouldn't this use the fringerprint instead? For me, `getprop ro.product.vendor.device` only prints out `vili`, which doesn't match this.
```suggestion
                android:requiredSystemPropertyName="ro.vendor.build.fingerprint"
```

PR #538, opened 2022-05-11
LG G7 One

AlexandreMarkus:
I would like to test it on my device. How could I test it?
LG G7 One overlay copied from LG G7
https://www.devicespecifications.com/en/model/37414c56


AlexandreMarkus:
followed this guide to do the fixes https://github.com/phhusson/treble_experimentations/wiki/How-to-create-an-overlay%3F

PR #539, opened 2022-05-11
[Moto One Fusion+] Update to stock 11 values

ksdrar:
Tested on v413

Fixes brightness at lowest level going black


phhusson:
Did you really need to change all of that, just to fix this issue?
You could have fixed this issue just by changing config_screenBrightnessSettingMinimum, no need to change the whole files...

ksdrar:
You're right, I just tested this with config_screenBrightnessSettingMinimum set to 2.

Should I close this PR and open a new one?

phhusson:
meh whatever. Just pick whichever you feel is safer

PR #541, opened 2022-05-11
[Nokia 8.1] Adjust rounded corner radius for Android 12L

TechnoStone:
Stop seeing white pieces of QS panel on lockscreen on light theme


TechnoStone:
Quick look on how it was:
example

PR #542, opened 2022-05-12
Sony pdx206: Use correct power_profile from stock

LumenCeleste:
I've experienced one completely weird bug with GSIs on my Xperia 5 II (pdx206) depending on battery, which have shown wrong capacities (something around 3000mAh and 3200mAh).

On hardware side, it's 4000mAh, but Sony seems to limit it to 3860mAh, maybe for battery protection. We should keep it as stock like as is. Also the cluster config was wrong because we use the 4-3-1 core scheme.

So this pr should implement the right power_profile for pdx206 devices.


DevZillion:
4000mAh is the Typical Capacity, Sony does not limit it to 3860mAh, that's the real capacity of it.

PR #543, opened 2022-05-12
Add overlay for Redmi10 2021 /Redmi Note11 4G (selene/selenes)

mochenya:


mochenya:
This is the effect of the overlay (the overlay in the vendor has been removed)
image

PR #550, opened 2022-06-01
Miscellaneous fixes for Sony Xperia 5 II (pdx206)

PeterCxy:
- Added an overlay for pdx206_kddi, the au (KDDI) variant of Sony Xperia 5 II. It shares the common configuration as the pdx206 target except the fingerprint matching rule.
- Enabled AOD support with an increased AOD brightness for legibility.
- A proper `power_profile` was imported. The previous one was just the dummy version shipped in AOSP by default.


PeterCxy:
The checks failed due to a missing `res` directory for pdx206_kddi, but I think it should be allowed in this context...

PR #552, opened 2022-06-03
M31s - Fix brightness issue on One UI 4 firmware

rufus582:
1. Fixed brightness issue on One UI 4 Firmware (by adding 2 configs from stock, that cause tests to fail)
2. Added System UI overlay to make screen on and off animations go towards the power button instead of volume

PS: The test will fail for sure because of the 2 lines below. But they are the ones that solve the brightness issue
```
    <item type="dimen" name="config_screenBrightnessSettingMaximumFloat">1.0</item>
    <item type="dimen" name="config_screenBrightnessSettingMinimumFloat">0.0</item>
```


Samsung/M31s/res/values/config.xml
```diff
@@ -586,32 +586,32 @@
...
-    </string-array>
     
     
     <bool name="config_fillMainBuiltInDisplayCutout">true</bool>
-    <bool name="config_hotswapCapable">true</bool>
     <bool name="config_showNavigationBar">true</bool>
-    <bool name="config_useDevInputEventForAudioJack">true</bool>
-    <bool name="config_wifi_dual_band_support">true</bool>
-    <bool name="config_wifi_fast_bss_transition_enabled">true</bool>
+    <bool name="config_supportAudioSourceUnprocessed">false</bool>
```
phhusson: I'm pretty confident that's wrong..

rufus582: Hmm🤔. Added those back now, and pushed that...

PR #553, opened 2022-06-07
Add Galaxy A33 resources

ChonDoit:


phhusson:
What's tested? What does this fix?

ChonDoit:
Auto Brightness + Extended one declared (detected max one is 420 by default)
No AoD on stock, so enabled doze/suspend and use Pulse for notifications
Sized and aligned status bar, declared cutout.
No led here.

Doze use prox ones is to avoid spam log outside of AOSP (Pocket Judge on somes)

Just forgot to delete proximity threshold, useless, still doesn't harm.

PR #563, opened 2022-06-13
Add overlay for Redmi Note 11S (Fixed)

IAGM12:
Basic overlay file for Redmi Note 11S

 Basically a copy of the Note 11 overlay but with correct values for our screen.


phhusson:
Is it tested?

IAGM12:
> Is it tested?

I tested on Linux. It's weird it's failing.

PR #570, opened 2022-06-20
overlay revised: add magnetic cover support for Xiaomi Pad 5 Pro 5G China

odys2061:
1. Add magnetic cover support from stock firmware's framework-res.apk(<bool name="config_lidControlsSleep">true</bool>)
2. unity of xml files under "~/Xiaomi/MiPad5Pro5G/res/values/" into single "config.xml"


phhusson:
Split that into two commits so i can actually see what you changed

odys2061:
I will reopen PR, now it's in a mess

PR #573, opened 2022-06-22
Added overlay for Xiaomi Poco X3 GT

jersoncarin:
1. Add support for Poco X3 GT Global for auto brightness + power service and other that has default from Poco X3 GT
2. Added  SystemUI overlay for the statusbar_height and rounded_corner _radius default


jersoncarin:
Why the build was cancelled? Anything error? I try on my Linux machine to tests but it's successful why here is failing?

phhusson:
I have no idea why. can you try pushing a new dumb commit to see what happens?

jersoncarin:
> I have no idea why. can you try pushing a new dumb commit to see what happens?

Sorry for that, I will try again thanks.

PR #580, opened 2022-06-27
Add overlay for Nokia 7.2

bulatorrr:
Auto brightness, Power profile, status bar padding, VoLTE. Tested on PE v414, Phh AOSP v414.


bulatorrr:
(OuO)

PR #578, opened 2022-06-24
Add overlay for Xiaomi Pad 5 (nabu)

AnzoP:
Used FrameworksResTarget.apk from the device.
Only relied on the guide as it's my first time creating a pull request.

PR #582, opened 2022-06-28
Add overlay for Xiaomi Pad 5

luisaosan:
Based on FrameworkResTarget.apk, extracted from device stock vendor.


phhusson:
Please discuss with #578

luisaosan:
As it seems that it's already wip i will close the pull request.

PR #587, opened 2022-07-03
Add overlay for Samsung Galaxy S20 Ultra z3s

quan3c10:


Samsung/z3s/res/values/config.xml
```diff
@@ -0,0 +1,256 @@
+<resources>
+    <string name="config_cameraLaunchGestureSensorStringType"/>
+    <string name="config_mainBuiltInDisplayCutout"/>
```
phhusson: uh? S20 Ultra definitely has a cutout

Samsung/z3s/res/xml/power_profile.xml
```diff
@@ -0,0 +1,67 @@
...
+        <value>400000</value>
+    </array>
+    <array name="cpu.active.cluster0">
+        <value>0.1</value>
+    </array>
+    <item name="cpu.idle">0.1</item>
+    <array name="memory.bandwidths">
+        <value>22.7</value>
+    </array>
+    <item name="battery.capacity">1000</item>
```
phhusson: that sounds doubtful

phhusson:
This overlay looks very wrong

Samsung/z3s/res/values/config.xml
quan3c10: It from strings.xml as below. I have run the test.sh and no fatal error exits.
image

Samsung/z3s/res/xml/power_profile.xml
quan3c10: I just see z3q device has the power_profile.xml file in xml folder, so just add this one from my device

quan3c10:
> 

What should i do next? just guide me how to check then i correct the overlay. Also please tell me how to add the overlay to "/system/product/overlay" since the system partition is read-only. I can double check the overlay before request a merge.

romanovj:
> > 
> 
> What should i do next? just guide me how to check then i correct the overlay. Also please tell me how to add the overlay to "/system/product/overlay" since the system partition is read-only. I can double check the overlay before request a merge.

push it to /vendor/overlay or use magisk module

quan3c10:
> > > 
> > 
> > 
> > What should i do next? just guide me how to check then i correct the overlay. Also please tell me how to add the overlay to "/system/product/overlay" since the system partition is read-only. I can double check the overlay before request a merge.
> 
> push it to /vendor/overlay or use magisk module

I have corrected the mentioned issue from @phhusson and tried your suggestion, however /vendor/overlay is read-only also. I tried with attached magisk module but it doesn't work. Could you please double check the magisk module is correct?
[add_overlay.zip](https://github.com/phhusson/vendor_hardware_overlay/files/9042609/add_overlay.zip)

quan3c10:
> > > > 
> > > 
> > > 
> > > What should i do next? just guide me how to check then i correct the overlay. Also please tell me how to add the overlay to "/system/product/overlay" since the system partition is read-only. I can double check the overlay before request a merge.
> > 
> > 
> > push it to /vendor/overlay or use magisk module
> 
> I have corrected the mentioned issue from @phhusson and tried your suggestion, however /vendor/overlay is read-only also. I tried with attached magisk module but it doesn't work. Could you please double check the magisk module is correct? [add_overlay.zip](https://github.com/phhusson/vendor_hardware_overlay/files/9042609/add_overlay.zip)

I made same config as z3q with z3s properties and tested. It work now

quan3c10:
@phhusson the overlay was tested on my device, Please merge the request so the z3s user won't need to use the magisk to make it work.

Samsung/z3s/res/xml/power_profile.xml
phhusson: So you're sating your device has a 1000mAh battery?

quan3c10: That my bad, I just took every number from framework-res.apk. How could I find all the correct number? They aren't mention in guideline :(

quan3c10: @phhusson I cloned the power_profile.xml from the z3s source trees on github, please take another look.

PR #592, opened 2022-07-10
Add overlay for XZ2

trinhloivn:
fixed headset with overlay


trinhloivn:
overlay for Sony XZ2

trinhloivn:
check error build

PR #593, opened 2022-07-10
Add overlay for Sony XZ2

trinhloivn:
Fixed headset with overlay


phhusson:
Is it tested? This does much more than just fixing headset

trinhloivn:
> Is it tested? This does much more than just fixing headset

yes i have tested it,it makes my headset work on build v414
It also makes the experience smoother than I thought
please merge it, thanks <3

PR #598, opened 2022-07-15
Fix overlays for Realme 8 5G

crazo7924:


crazo7924:
@phhusson now the overlays work as expected after boot. 
Manifest pattern matching is weird

phhusson:
Is it tested ok?

crazo7924:
Yes and I am using it with PixelExperience plus

PR #602, opened 2022-07-17
Add overlay for Xiaomi Poco M4 Pro 5G

mmtrt:


Xiaomi/PocoM4Pro5G/res/xml/power_profile.xml
```diff
@@ -0,0 +1,138 @@
+<?xml version="1.0" encoding="utf-8"?>
+<device name="Android">
+    <item name="ambient.on">0.1</item>
```
phhusson: With a IPS screen? That's doubtful

phhusson:
Is it tested?

Xiaomi/PocoM4Pro5G/res/xml/power_profile.xml
mmtrt: I couldn't find proper power profile so I just kanged from Poco m3 PR though the one I found were dummies not sure where to look.

mmtrt:
Testing more changes will let you know.

mmtrt:
Tested and works fine.
[padding_overlay.zip](https://github.com/phhusson/vendor_hardware_overlay/files/9219558/padding_overlay.zip)

rusantiman:
Hi
I respect you guys because I have no knowledge.

I put it in overlay.
It is an impression that I tried.

The problem is
Auto brightness is turned off and cleared by restarting.
There is no particular difference from the state without this overlay.
I think the stock default is strange.

It was notch kill in full screen.
I think it is better to additionally include the following.

string name = "config_mainBuiltInDisplayCutout"> M 0,0 H -37 V 96 H 37 V 0 H 0 Z </ string

It was in miui 13 DeviceAndroidConfig / strings.xml.

for your information

mmtrt:
Already tested that string config, didn't include b/c it cuts the notch path when the device is in landscape.

rusantiman:
OK. If you knew it, there is no problem.
It was different from the defaults of some apps, but for me it has no particular effect.
The photo below is normal.

Is there a problem with auto brightness?



Screenshot_20220722-085016_ChMate
Screenshot_20220722-083221_ChMate
.

rusantiman:
Sorry
I made a mistake in English.
Anyway, automatic brightness isn't working well.
"Force alternative backlight scale" is on.
I think the stock default settings is strange.

I have no knowledge, so copy the overlay settings of my nokia 7.1.
I am changing it irresponsibly.
I'm using it because it can't be helpe.

AOSP v414　global miui13

mmtrt:
@rusantiman test this for notch.
[padding_overlay.zip](https://github.com/phhusson/vendor_hardware_overlay/files/9165412/padding_overlay.zip)

rusantiman:
No problem.
I tried it with some apps.
Thank you.

PR #610, opened 2022-07-31
Add wifi mac randomization for Samsung Galaxy A53

aximut:
Setting **not tested** on real device yet, but standard root/adb MAC address change works, so no reason to not include this for more privacy (unless this setting is known to break things).

### DIFF
--- a/Samsung/A53/res/values/configs.xml
+++ b/Samsung/A53/res/values/configs.xml
@@ -495,7 +495,9 @@
     <bool name="config_device_volte_available">true</bool>
     <bool name="config_fillMainBuiltInDisplayCutout">false</bool>
     <bool name="config_showNavigationBar">true</bool>
+    <bool name="config_wifi_connected_mac_randomization_supported">true</bool>
     <bool name="config_wifi_dual_band_support">true</bool>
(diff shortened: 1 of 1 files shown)


aximut:
If there is a build pipeline and I could get the resulting overlay apk, I could test it using Magisk.

aximut:
I tested this as an overlay using Magisk ([similar to this one](https://github.com/diogotcorreia/treble-overlay-xiaomi-mi11tpro)), but the "Privacy" option still does not show up in WiFi settings. Does anybody know why?

The mac address change using `ifconfig` works on the device.

aximut:
WARNING
Don't merge, touching MAC address will likely corrupt your `efs` and `sec_efs` partition (BACK THEM UP BEFORE DOING ANYTHING LIKE THIS!!!). In particular, the `nv_data.bin` gets corrupted and you will lose IMEI and GPS (painful!).
Samsung makes it hard to implement a feature like this.
See also [this thread](https://forum.xda-developers.com/t/tutorial-recover-from-imei-or-gps-not-working-due-to-corruption-restoring-efs-and-sec_efs.4476689/).

PR #628, opened 2022-08-27
Add overlay for Nokia 3.2 (Deadpool)

thomas-catt:
(not particularly bug-free, adaptive brightness failed for me, but since notch, idle drain, and battery stats work now, its worth a pull)


phhusson:
Please commit again without marking all files as executables in the process.

PR #631, opened 2022-08-29
feat: add notch & update values from latest a12 framework of Redmi 10X 5G (atom)

Deivu:
The changes are all tested with my own Redmi 10X 5G

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -1,7 +1,7 @@
 name: Overlay
 
 on:
-  push:
+  pull_request:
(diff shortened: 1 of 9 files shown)


phhusson:
I guess the code name is actually "atom"?

Deivu:
Yes, this device codename is atom

Deivu:
Hello, I added new commits since my pr, so while I wait, I decided to update the values based on latest builds of the stock rom of this phone. This changes **are all tested and confirmed working** on my own phone

PR #645, opened 2022-09-20
Update pinner list locations for Redmi 9

ghost:


phhusson:
It's better to remove it rather than setting empty if ROM defines its own pinner list.

Also though it's obsolete why remove it? You found issues because of that?

cangkurileung:
Just did an update @phhusson

cangkurileung:
Adjust files locations @phhusson

PR #680, opened 2022-12-15
Update Redmi 9 overlays

ghost:
Sir @phhusson 
The density of Redmi 9 is 440
So i removed the unused values.


phhusson:
User can change DPI in Settings though

ghost:
> User can change DPI in Settings though

Ah ok.
But sometimes this happens https://t.me/phhtreble/683813

phhusson:
How is that related?

Le jeu. 15 déc. 2022 à 11:31, cangkurileung ***@***.***> a
écrit :

ghost:
idk, i just wanted to make it look clean.

```
<dimen name="rounded_corner_content_padding">4.5dip</dimen>
<dimen name="rounded_corner_content_padding">7.0dip</dimen>
```
they had no effect so I chose to delete them.

but it's okay if you don't think it's necessary.

phhusson:
I think they need to be here, to correct values. Try changing ""screen size"" in  Settings => Display and check  that the UI is still correct

ghost:
> I think they need to be here, to correct values. Try changing ""screen size"" in Settings => Display and check that the UI is still correct

Sir @phhusson sorry for late response.

Unlike R vendor , S vendor doesn't load systemui because they moved DevicesOverlay.apk (which is responsible for UI) from /system/overlay/ to /product/overlay/
So I made a Redmi9-SystemUI overlay with 3 folders (from stock MIUI).

1. **values-440dpi** with `<dimen name="rounded_corner_content_padding">19.0dip</dimen>`
2. **values-xxhdpi** with `<dimen name="rounded_corner_content_padding">7.0dip</dimen>`
3. **values** with `<dimen name="rounded_corner_content_padding">4.5dip</dimen>`

I deleted the **values-xxhdpi** & **values** folder because the correct one is **values-440dpi**
**values-xxhdpi** & **values** are wrong because in the status bar the clock and battery positions are not balanced.

i checked using `adb shell getprop | grep density` & `adb shell dumpsys display`
the result is `[ro.sf.lcd_density]: [440]` & `density 440`

***
Here is the comparison:

### values-440dpi (the correct one)
Screenshot_20221215-230939
Screenshot_20221215-230947
Screenshot_20221215-230957
Screenshot_20221215-231008

***
### values-xxhdpi
Screenshot_20221215-232443
Screenshot_20221215-232450
Screenshot_20221215-232458
Screenshot_20221215-232513

***
### values
Screenshot_20221215-231827
Screenshot_20221215-231835
Screenshot_20221215-231843
Screenshot_20221215-231852

phhusson:
`ro.sf.lcd_density` only contains default/main density. When  changing "Display size" setting it does change dpi, you can see it with `adb shell wm density`

Anyway your screenshot say that current overlay is just fine over all display sizes, so there  is no need  to change it?

ghost:
> `ro.sf.lcd_density` only contains default/main density. When changing "Display size" setting it does change dpi, you can see it with `adb shell wm density`
> 
> Anyway your screenshot say that current overlay is just fine over all display sizes, so there is no need to change it?

Oh, my bad.

No, the current overlay is not fine.
The screenshot above is when i test the values one by one.

Current overlay contains 3 folders/values.
If i change the display size, the status bar goes wrong (at densities 490 and 540)
So all I need is **values-440dpi** to make the status bar position stable.
And delete the rest.

ghost:
Hallo again sir @phhusson 

The culprit has been found -> **values-xxhdpi**
Without it the status bar is fine.
I also added `navigation_side_padding` & `status_bar_padding_top`

Tested.
Thank you!

PR #691, opened 2022-12-27
FP3: Enable night mode

danielzhang130:
The stock OS sets this value to true


phhusson:
Who the duck set it to false.... It's enabled on all devices...

danielzhang130:
i believe it was set to false on /e/foundation, so it's being copied over.

PR #692, opened 2022-12-27
FP3: Enable fingerprint sensor

danielzhang130:


phhusson:
It's already enabled on all devices

danielzhang130:
@phhusson Thanks for the review. I don't believe it's enabled on FP3. There is no fingerprint option in settings. This patch defines the available sensors. 

I also have a pixel 3 which sets this value via an overlay in the vendor directory, so I believe without that, fingerprint wouldn't work on pixel 3 neither. FP3 doesn't have that overlay in vendor, so it doesn't have fingerprint working. 

I'm building on trebledroid 13 btw, would stuff be different somehow?

phhusson:
Which TrebleDroid 13? On which android vendor version? At some points Android vendors <= 10 were missing fingerprint but that was fixed in https://github.com/TrebleDroid/platform_frameworks_base/commit/15a193cf0b49241d957c36393bfb79b75c97135e

danielzhang130:
I'm on https://github.com/TrebleDroid/treble_manifest/commit/6933f4e6959c0a4cd96c91b48d68ffe354574737

If it helps, I found this on /e/foundation as well https://gitlab.e.foundation/e/devices/android_device_fairphone_FP3/-/commit/5a9506dae9ffb7131e2918599380cdddd4b92c0d

PR #696, opened 2023-01-01
Update overlays for galaxy A32 and add system ui overlays.

realahnet:
I made new system ui overlays that actually work! and new configs. Tested the overlays out. they work great.

### DIFF
--- a/Samsung/A32-SystemUI/res/values/bools.xml
+++ b/Samsung/A32-SystemUI/res/values/bools.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <bool name="config_enableWifiDisplay">true</bool>
+    <bool name="config_adaptive_display_solution_enabled">true</bool>
+    <bool name="config_all_around_sensing_enabled">false</bool>
+    <bool name="config_big_data_logging_enabled">false</bool>
+    <bool name="config_dynamic_automatic_brightness_available">true</bool>
+    <bool name="config_eye_comfort_solution_enabled">true</bool>
+    <bool name="config_mdnie_scenario_control_enabled">true</bool>
+</resources>
--- a/Samsung/A32-SystemUI/res/values/bools.xml
+++ b/Samsung/A32-SystemUI/res/values/bools.xml
@@ -1,10 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <bool name="config_enableWifiDisplay">true</bool>
-    <bool name="config_adaptive_display_solution_enabled">true</bool>
-    <bool name="config_all_around_sensing_enabled">false</bool>
-    <bool name="config_big_data_logging_enabled">false</bool>
-    <bool name="config_dynamic_automatic_brightness_available">true</bool>
     <bool name="config_eye_comfort_solution_enabled">true</bool>
     <bool name="config_mdnie_scenario_control_enabled">true</bool>
 </resources>
--- a/Samsung/A32-SystemUI/res/values/dimens.xml
+++ b/Samsung/A32-SystemUI/res/values/dimens.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <item type="dimen" name="config_screenBrightnessDimFloat">0.047058824</item>
+    <item type="dimen" name="config_screenBrightnessSettingDefaultFloat">0.5019608</item>
+    <item type="dimen" name="config_screenBrightnessSettingMaximumFloat">1.0</item>
+    <item type="dimen" name="config_screenBrightnessSettingMinimumFloat">0.0</item>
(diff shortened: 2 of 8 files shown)


SysWhiteDev:
how did you check if the overlay was working? because i have no idea how

realahnet:
> how did you check if the overlay was working? because i have no idea how

i pushed the built apks to my /system/product they seemed to be working and i didnt see any errors in adb logcat. i didnt see any errors about the extra properties that do not exist. so they work jsut fine.

SysWhiteDev:
sorry if i wasn't clear, but i mean what should the phone do differently if it is really working?

realahnet:
> sorry if i wasn't clear, but i mean what should the phone do differently if it is really working?

i fixed some values which did make an effect. the brightness slider was working way more accurately. i also added a config for biometric sensors because on new td patches fp sensor was being disabled for most users because it allowed pixel devices to boot. this is to compensate for that. i would like to have that in there incase something like this happens again. this fixed it on ricedroid gsi. so there are some changes and system ui overlay also helped! that status bar potrait config was also fixed some wierd placement there. so yes i did not just add unncessary configs. they all help.

realahnet:
one more thing. just noticed pushing my new overlays into some gsi's fixes face unlock somehow.

Samsung/A32-SystemUI/res/values/dimens.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <item type="dimen" name="config_screenBrightnessDimFloat">0.047058824</item>
```
phhusson: that's not a sysui prop...

Samsung/A32-SystemUI/res/values/bools.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <bool name="config_enableWifiDisplay">true</bool>
+    <bool name="config_adaptive_display_solution_enabled">true</bool>
+    <bool name="config_all_around_sensing_enabled">false</bool>
+    <bool name="config_big_data_logging_enabled">false</bool>
+    <bool name="config_dynamic_automatic_brightness_available">true</bool>
```
phhusson: that's not in systemui

phhusson (changes requested):
You're confusing systemui and frameworks-res overlay...

Samsung/A32-SystemUI/res/values/dimens.xml
realahnet: im sorry i wasnt aware. i am not an expert with overlays. i'll remove them

Samsung/A32-SystemUI/res/values/bools.xml
realahnet: okay i'll remove those. sorry i am not an expert in overlays

Samsung/A32-SystemUI/res/values/dimens.xml
realahnet: configs removed

Samsung/A32-SystemUI/res/values/bools.xml
realahnet: configs removed

Samsung/A32-SystemUI/res/values/bools.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
```
phhusson, line 2: none of those bools are in sysui

Samsung/A32-SystemUI/res/values/dimens.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <item type="dimen" name="config_screenBrightnessSettingDefaultFloat">0.5019608</item>
```
phhusson, line 3: none of the config_xxxx are in sysui

phhusson (changes requested):

Samsung/A32-SystemUI/res/values/bools.xml
realahnet: i dont understand. does that mean my whole system ui overlay is useless?

Samsung/A32-SystemUI/res/values/dimens.xml
realahnet: so that means my whole system ui config is useless?

phhusson: no you can keep the one not starting with config_

realahnet: oh okay. btw i am not exprienced with overlays. but is there a list of configs or something where i could know what all the configs are? i followed the guide and thats how i made these 2 overlays.

PR #698, opened 2023-01-08
Add Overlay for Redmi Note 11t 5G

weaponmasterjax:
Tested and should work fine.


weaponmasterjax:
@phhusson
sorry to ping you but please have a check?

PR #699, opened 2023-01-09
[Minor] Nord N10 Priority adjust

FinecoFinit:
Priority was too low; overlay was overwritten by other and wasn't able to change status bar


FinecoFinit:
@phhusson ready for implementation

PR #705, opened 2023-01-23
Add Umidigi A13 Pro Max 5G overlays

JonnyVR1:


Umidigi/A13_Pro_Max_5G/res/xml/power_profile.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="utf-8"?>
+<device name="Android">
+    <item name="ambient.on">0.1</item>
+    <item name="screen.on">0.1</item>
```
phhusson, line 4: that's very doubtful. This looks like sample power_profile

Umidigi/A13_Pro_Max_5G/res/values/integers.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <integer name="config_screenBrightnessSettingMaximum">255</integer>
+    <integer name="config_screenBrightnessSettingMinimum">3</integer>
+	<integer name="config_defaultPeakRefreshRate">61</integer>
+    <integer name="config_defaultRefreshRate">61</integer>
```
phhusson, line 6: What?

phhusson (changes requested):
This power profile looks completely wrong. It looks like it's been taken from AOSP's example power profile except that you changed battery capacity.
I don't understand wtf that 61Hz mode

Umidigi/A13_Pro_Max_5G/res/values/integers.xml
JonnyVR1: From the vendor overlay of framework-res
image

Umidigi/A13_Pro_Max_5G/res/xml/power_profile.xml
JonnyVR1: From the vendor overlay
image

JonnyVR1: FYI this is consistent with other Umidigi overlays in this very repo - https://github.com/phhusson/vendor_hardware_overlay/blob/pie/Umidigi/A5pro/res/xml/power_profile.xml

phhusson: well that means you didn't take correct overlay (or you shouldn't take the overlay but framework-res), or umidigi'spower profile is broken.
And ok i'll remove the A5  Pro one

JonnyVR1: I don't think removing it is the correct solution here - if you run a notepad++ scan on all the power profile files, there are many occurrences of these values across different vendors so to me this looks like the power profile submitted in this PR and obtained directly from the stock ROM vendor overlay is correct:

image

phhusson: There isn't any smartphone screen that consumes 0.1mA when on.

I'll improve the tests so that I stop merging overlays with such stupid profiles, thanks for bringing that up

JonnyVR1: In that case, do you know of how I would go about obtaining the correct values?

phhusson: do `grep power_profile -r /system/frameworks /vendor/overlay /product/overlay /odm/overlay`, and decompile all the APKs that are listed by that command, and see if there is one with values that make more sense.

JonnyVR1: None of the matching files have values that make any more sense - the one in /vendor/overlay is what I took the values from, /product/overlay has the same values

image

phhusson: what about the one in system/framework/framework-res.apk?

JonnyVR1: framework-res.apk contains these files:

power_profile.xml
power_profile_test.xml

power_profile.xml has the same values but without additional CPU clusters and an incorrect battery capacity - I think this is the aosp template one:
image

power_profile_test.xml has what seems like more sensible values but I'm not sure if this file is used, and it has an incorrect battery capacity:
image

phhusson: K great, let's just stick to the least broken power_profile of the ROM (no the _test is not ok)
The one in /product/overlay/xxxx has the correct battery capacity or did you have to change?
If you didn't have to change it, let's take that.
All measurements of power consumption will be totally broken, but i guess it won't be more broken that stock...

JonnyVR1: The one in /product/overlay had an incorrect battery capacity, the original one submitted in the PR (/vendor/overlay) had the correct one, I didn't have to change it

phhusson: Would you be able to tell whether stock ROM reports non-zero power consumption from the display?

JonnyVR1: I'm not sure how to do that

PR #720, opened 2023-02-23
Remove "Add overlay for ONN 7 Tablet 3rd Generation"

jquagga:
This is the tablet build which is currently breaking the overall build.  This hopefully simply reverts the commit and removes the overlay for now getting things back to building.


phhusson:
Nope, I won't remove it. Just rename it.

PR #721, opened 2023-02-23
Rework S21 Ultra (p3s) overlays

TQMatvey:
TLDR: add proper AOD, DOZE state, Animation on/off AOD transitions, status bar icons are now not going in edges, burnin protection (tested, works), add comments to everything, other stuff


TQMatvey (commented):
Due to "doze_display_state_supported"=true
after some time in "aod" (2-3 mins) the AOD freezes completely
trying to turn the phone on works after few attempts but the phone will be running super slow until quickly turning off and on
removing doze_display_state_supported fixes the issue addressed above
but it doesnt doze and burnin protection isnt being initialized.

TQMatvey:
upon removing doze_display_state_supported display doesnt start dozing `mScreenState=ON`
looking at the Burnin protection code we can see that it expects display to be in STATE_DOZE, STATE_DOZE_SUSPEND or STATE_ON_SUSPEND

enabling doze_display_state_supported sets display to `mScreenState=DOZE` and therefore Burnin protection activates

reference: https://katb.in/anelikotuju

TQMatvey:
just rebooted, issue is gone... this definately needs more testing, very weird

UPD: died again

TQMatvey:
in current state it seems to work fine, it still sometimes shows little temporary (for less than 30s) burnins, overall everything is quite good

phhusson:
`Fatal: ./Samsung/p3s/res/values/config.xml: defines a non-existing attribute config_udfps_sensor_props`
do you need it? it shouldn't be needed

TQMatvey:
> `Fatal: ./Samsung/p3s/res/values/config.xml: defines a non-existing attribute config_udfps_sensor_props` do you need it? it shouldn't be needed

indeed, its not **required** but its **preferred**, it make FOD a bit smaller, gives it more "Pixel" look, and overall looks better on screen and doesnt overlap the large lockscreen clock

TQMatvey:
tho if that warning is a bit annoying, i can remove that, its not critical.

phhusson:
Ok I understand why you would want that, though you're making the sensor "lower quality" by doing it (having a bigger surface for FOD is a feature so that the user can be less precise in its finger position).

I'm fine merging that in. If you still want it, rebase on latest commit, and add a comment in that xml saying "I swear it makes sense to set config_udfps_sensor_props, and I can completely explain why."

TQMatvey:
uhm i am really sorry, i have no clue on how to rebase, i tried googling but its a topic i just cant figure out...

TQMatvey:
i just did a test, and it isnt that bad to remove those props (i remember it being way worse)

PR #722, opened 2023-02-25
Add Samsung Galaxy S20 Ultra Exynos (z3s)

Nico170420:


phhusson:
power_profile is obviously wrong

PR #724, opened 2023-03-03
Add auto brightness values for Asus Rog Phone (Z01QD)

Rocky7842:


phhusson:
`<integer name="config_screenBrightnessDoze">17</integer>`
this sounds extremely high and power hungry to me, but well, i'll trust your judgement

PR #728, opened 2023-03-04
New overlay for MAR

Iceows:
MAR is the codename for Huawei P30 lite - kirin 710


phhusson:
I can merge that if you want, but power_profile is obviously wrong.

Iceows:
I look, it seems to me that I took the file of the android 9 package which comes from Huawei. I confess that I have not looked more than that

PR #729, opened 2023-03-06
Rename ONN 100071481 to mid7019 to hopefully fix build

jquagga:
Ok, this is a renaming attempt.  I've left the systemPropertyValue below there which is numeric but hopefully won't cause an issue.  The .apk builts and passes test.sh.  I haven't tried building a full with this yet but that's in progress.  

I've left android:requiredSystemPropertyValue="+*onn/100071481*" with the numeric string but I've changed the description name to be not entirly numeric.  I'm using the tablet's "other name" based on below:

[ro.bootimage.build.fingerprint]: [onn/100071481/mid7019_mr_32:11/RP1A.200720.011/qyh20220216:user/release-keys]


phhusson:
Please test it. I'm pretty sure it doesn't build
As far as I understand the issue is exclusively with packageName= in AndroidManifest.xml

jquagga:
Ok, I am happy to report that it now builds through the build.sh process after that second commit.

phhusson:

PR #735, opened 2023-03-20
Add overlay for Duoqin Qin F21 Pro

Saphar:


DuoQin/QinF21Pro/AndroidManifest.xml
```diff
@@ -0,0 +1,10 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+        package="me.phh.treble.overlay.duoqin.qinf21pro"
+        android:versionCode="1"
+        android:versionName="1.0">
+        <overlay android:targetPackage="android"
+                android:requiredSystemPropertyName="ro.vendor.build.fingerprint"
+		android:requiredSystemPropertyValue="+Qin/full_k61v1_64_gms*"
```
phhusson, line 7: that doesn't sound like a great fingerprint match, but then there are probably not many qin devices

PR #736, opened 2023-03-21
Add the Overlay for Huawei JKM

Ryker0603:
Add the Overlay for Huawei JKM, the test model is JKM-LX2, it works well. Other models should work as long as they are JKM.


phhusson:
You didn't list the overlay in overlay.mk

PR #738, opened 2023-03-23
Lenovo Legion Y70 Overlay - added missing values from FrameworkResTarget_vendor.apk

MmmmmDonuts:
Initially only framework-res.apk and frameworkrescommon.apk were included.
Added missing values from FrameworkResTarget_vendor.apk

At the very least, this will fix adaptive brightness with the line:
bool name="config_automatic_brightness_available" true


phhusson:
you don't need to copy/paste everything from every apk you find on earth, you can just set the values you need...

MmmmmDonuts:
Following testing, the bool name="config_automatic_brightness_available" true line does add the option for automatic brightness in settings. However, it doesn't actually do anything. brightness remains fixed. Further troubleshooting is required. À suivre...

PR #739, opened 2023-03-27
[Lenovo Y70] Add fingerprint gesture support

MmmmmDonuts:


phhusson:
Is it tested and it works?

MmmmmDonuts:
Finally got around to testing. Unfortunately it doesn't work. The setting briefly appears and then disappears in gestures.

PR #741, opened 2023-04-07
Update Galaxy A32 4G overlays

realahnet:
I made changes to fix AOD on this device. which fixes the always on display getting stuck and not updating. I have also realigned the power button for the animation between AOD and lock screen. It works great. also enabled some misc configs and added burn in protection. All tested and working.


realahnet:
reopening in updated td repo.

PR #743, opened 2023-04-15
Add overlay for Samsung Galaxy A02s

mrugtangy:


mrugtangy:
I removed a device by mistake in overlay.mk in the first commit and added it back in the second one.

phhusson:
That's a very light power profile. Did you check that it works at least a bit?

tangymc:
> That's a very light power profile. Did you check that it works at least a bit?

Yep it does work.

mrugtangy:
> That's a very light power profile. Did you check that it works at least a bit?

Works perfectly. And I also found it in the device's framework-res.apk (vendor) and it contains the right configs for the device.

mrugtangy:
Pls pull 🙂

PR #766, opened 2023-10-13
Add overlay for Tecno Spark 9Pro

N3Shemmy3:
Hi i created an overlay for Tecno  Spark 9Pro using your guide and was hoping to add it to the phh overlays


N3Shemmy3:
i had fogotten that my device is the pro vairiant so ive made changes and yay it builds!

PR #768, opened 2023-10-17
Add overlay for Xiaomi 11T

ziadtamer11:


ziadtamer11:
Add overlay for Xiaomi 11T

PR #770, opened 2023-10-23
Add overlay for Poco F4 GT (`ingres`)

ghost:
I'm not an android expert by any mean.

I followed instructions on [How to create an overlay? page](https://github.com/phhusson/treble_experimentations/wiki/How-to-create-an-overlay%3F) and tested the `.apk` file on my device (thanks to `magisk-overlayfs` module) and everything is working good so far!

The only thing I had the chance to test was autobrightness and no issues on that feature.


phhusson:
Remove power_profile_test. power_profile is obviously wrong.
If you only tested autobrightness, then only include values related to auto-brightness.

phhusson:
Ah also you're in the wrong place, the correct place is https://github.com/TrebleDroid/vendor_hardware_overlay

PR #790, opened 2025-02-01
qin f21 pro non-gms also need to be supported.

yunshaoyy:


phhusson:
commit is good, but wrong git. Correct git is https://github.com/trebledroid/vendor_hardware_overlay/

yunshaoyy:
ok,thanks for your reply.

